PR #4, opened 2021-11-17
Draft STPA for stack protection

reiterative:


reiterative:
@coppermanME Thanks for the input, Elana! Sorry that I forgot to include a link to this PR - please can you add your comments here instead of in the files?

coppermanME:
[EC] Why only stack overflow issues?
[EC] Why only data protection?  Code may also be corrupted by stack overflow.

[EC] Why is stack depth overflow a concern only for kernel data structures, whereas stack overflow is a concern only for application data?


[EC] Assets to be protected are data and code, in kernel and user space.

reiterative:
> [EC] Why only stack overflow issues?

This was just a quick initial draft. Very happy to add other hazards related to the stack!

> [EC] Why only data protection? Code may also be corrupted by stack overflow.

OK. I'll add that.

> [EC] Why is stack depth overflow a concern only for kernel data structures, whereas stack overflow is a concern only for application data?

I was reading about stack depth overflow in the kernel context, and hadn't made the logical connection that it might apply in userspace too. 

> [EC] Assets to be protected are data and code, in kernel and user space.

Will update.

PR #11, opened 2022-08-22
Pb glossary additions

reiterative:


framework/framework.md
```diff
@@ -24,6 +24,39 @@ The following terms are used in this document with a specific meaning. Definitio
...
+* **assessment**: The examination of whether or not a *system* achieves the safety objectives.
+* **audit**: The examination of an implemented process with regard to the process objectives.
+* **baseline**: The version of an approved set of one or more work products that serves as a basis for change.
+* **component**: A non-system level *element* that is logically or technically separable and is comprised of more than one hardware part or one or more software units. (TODO - Normalize this definition with the component above)
+* **configuration data**: The data that is assigned during element build and that controls the element build process. (TODO - Reword to make more comprehensible)
+* **element**: A system, hardware or software component, hardware part, or software unit.
+* **error**: The discrepancy between a computed, observed or measured value or condition, and the true, specified or theoretically correct value or condition.
+* **failure**: The termination of an intended behaviour of an element due to the manifestation of a fault.
+* **fault**: An abnormal termination that can cause an *element* to fail.
```
reiterative, line 38: I think this should be "An abnormal *condition*..."

framework/framework.md
```diff
@@ -24,6 +24,39 @@ The following terms are used in this document with a specific meaning. Definitio
...
+* **fault tolerance**: The ability to deliver a specified functionality in the presence of one or more specified *faults*.
+* **fault tolerant time interval**: The minimum time-span from the occurrence of a *fault* to the possible occurrence of a hazardous event, if the safety mechanisms are not activated. (TODO -- Is there a better or more generic terminology for this?)
+* **inspection**: The examination of work products, following a formal procedure, to detect safety anomalies.
+* **latent fault**: A fault whose presence is not detected by a safety mechanism nor perceived by the driver within an appropriate time.
```
reiterative, line 42: References *safety mechanism*, which isn't defined

framework/framework.md
```diff
@@ -24,6 +24,39 @@ The following terms are used in this document with a specific meaning. Definitio
...
+* **malfunctioning behavior**: A *failure* or unintended behavior of an *element* with respect to its design intent.
+* **modified condition/decision coverage**: (MC/DC) A code coverage metric that indicates the percentage of single condition outcomes that independently affect a decision's outcome.  In addition, it expects that all code blocks and execution paths have been tested.
```
reiterative, line 44: I'm not convinced that it is necessary to include this, as it is a specific and unambiguous term

framework/framework.md
```diff
@@ -24,6 +24,39 @@ The following terms are used in this document with a specific meaning. Definitio
 * **target equipment**: [[IEC]](https://std.iec.ch/terms/terms.nsf/0/B9C576C31B4E488DC125771F002F58E9?OpenDocument) the equipment, machinery, apparatus or plant for which *safety* is to be achieved
 * **target system**: The *safety-related system* (specific or assumed) under consideration. For the purposes of this document, this is assumed to include a *component* incorporating software based on *Linux*
 
+### ISO 26262 Definitions
+
+* **architecture**: The representation of the structure of the *system* that allows identification of building blocks, their boundaries and interfaces, and includes the allocation of requirements to those building blocks.
+* **assessment**: The examination of whether or not a *system* achieves the safety objectives.
```
reiterative, line 30: I feel this should probably be 'safety assessment'

framework/framework.md
```diff
@@ -24,6 +24,39 @@ The following terms are used in this document with a specific meaning. Definitio
 * **target equipment**: [[IEC]](https://std.iec.ch/terms/terms.nsf/0/B9C576C31B4E488DC125771F002F58E9?OpenDocument) the equipment, machinery, apparatus or plant for which *safety* is to be achieved
 * **target system**: The *safety-related system* (specific or assumed) under consideration. For the purposes of this document, this is assumed to include a *component* incorporating software based on *Linux*
 
+### ISO 26262 Definitions
+
+* **architecture**: The representation of the structure of the *system* that allows identification of building blocks, their boundaries and interfaces, and includes the allocation of requirements to those building blocks.
```
reiterative, line 29: I feel this should be 'system architecture' - we might also have architecture for an element or component

framework/framework.md
```diff
@@ -24,6 +24,39 @@ The following terms are used in this document with a specific meaning. Definitio
...
+* **software unit**: An atomic-level software *component* of the software architecture that can be subjected to stand-alone testing.
+* **system**: A set of *components* or subsystems that relates at least a sensor, a controller, and an actuator with one another. (TODO -- Normalize with the above definition)
+* **systematic failure**: A *failure* related in a deterministic way to a certain cause, that can only be eliminated by a change of the design or of the manufacturing process, operational procedures, documentation or other relevant factors.
+* **systematic fault**: A *fault* whose *failure* is manifested in a deterministic way that can only be prevented by applying process or design measures.
+* **testing**: The process of planning, preparing, and operating or exercising an *element* to verify that it satisfies specified requirements, to detect *safety anomalies*, to validate that requirements are suitable in the given context and to create confidence in its behavior.
+* **verification**: The determination whether or not an examined object meets its specified requirements.
```
reiterative, line 57: We may need a definition of *validation* as well (which is sadly lacking in ISO 26262)...

reiterative (commented):
I think many of these definitions are helpful, and provide a broader context for what we need to consider when we are thinking about safety. I especially like the 'testing' and 'safety measure' definitions.

On the other hand, some of them feel unhelpfully narrow in their definition, particularly where they define (or relate to) ISO 26262's concepts of 'system', 'component' and 'element'. Does the meaning of the related terms change significantly if we use my definitions of these (and substitute 'component' for 'element')?

PR #17, opened 2023-04-20
Document safety analysis approach

reiterative:
Describing OSEP's proposed approach to safety analysis for Linux, for sharing with the wider ELISA (and open source ) community.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,8 @@ collaborative work-in-progress (topic suggestions, notes, etc) in the project
 ## Project index
 
 * [Mission statement](mission.md)
-
+* [Safety analysis approach](safety-analysis/README.md)
+* [Meeting minutes (*Wiki*)](https://github.com/elisa-tech/wg-osep/wiki/minutes)
 ## External links
 
 * [ELISA home page](https://elisa.tech/)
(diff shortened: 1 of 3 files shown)


igor-stoppa:
I am not proposing changes because I am not sure what woule be the best way to phrase hte thoughts below, but here comes nothing:

- I would want to have further point that details how linux - as it is now - might already support some f those goals
- similarly, I would want to have also a list of what impediments linux might introduce in meeting those goals; in this regard, I think it's important to differentiate between gaps and real impediments, where the linux design actively goes against safety goals, because gaps can be filled, but if something is designed in a certain way, which happens to hamper safety, then it is necessary to introduce mitigations that do not jeopardize the *other* non safety goals that such component was supporting. Someone might decide to de-prioritize them, as trade-off for improved safety, but then it should be clear what is being given up.

reiterative:
Thanks for your inputs @igor-stoppa.

Please check the updated text and confirm that we have addressed your points, and [review / approve](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/approving-a-pull-request-with-required-reviews) if you are happy.

reiterative:
Updated to move the approach text into its own directory, where we can begin to document both the steps for applying the approach and our results.

kent-ul (approved):
Review complete; no comments.

PR #21, opened 2023-10-13
add first draft for PiU paper

shetze:
as discussed in our meeting yesterday, here is the first draft as PR

### DIFF
--- a/Proven-In-Use-Argument.md
+++ b/Proven-In-Use-Argument.md
@@ -1,0 +1,77 @@
+# Is Linux mature and proven enough for use in safety-critical applications?
+
+Linux has proven itself thousands, probably millions of times over in business-critical applications. So it makes sense to use Linux for safety-critical applications as well. Linux has many advantages 
+
+* it runs on an enormously wide range of hardware
+* there are a large number of established vendors and service providers for the operating system and for additional (standard) software
+* many specialists are familiar with it
+* it is continuously being developed
+* a large part of the innovation in IT is created on Linux and as open source software
+* ...
+
+Linux has obviously proven itself. So what is so difficult about safety-critical applications?
+
+First, it's useful to look briefly at what functional safety is actually about. At its core, it is about dangerous items: technical systems, devices, machines, vehicles and the like. Such items can harm the environment or people, and in the worst case can even kill people uninvolved in their operation.
+
+For obvious reasons, the dangers posed by an item must be reduced to an acceptable level before the item is put into operation or placed on the market. 
+Thus, to reduce the unavoidable risks from an item to an acceptable level, appropriate safeguards must be found and implemented at the design stage. These protective measures come into action under certain conditions and regulate the item in such a way that a previously identified and determined risk remains below a predetermined threshold. These protective measures are called safety functions. In certain situations, a safety function can put the item in a state recognized as safe, for example, by shutting the item down.
+
+With the introduction of safety functions, the risk posed by an item now depends on the correct operation of that protective measure. Thus, the item is safe only to the extent that the safety function operates correctly and reliably. This is the safety of the safety function, or simply functional safety.
+
+It is obvious to implement safety functions at least partially with electronic or programmable components. It is only in this context that the question of using Linux for safety-critical applications arises: as an operating system for a programmable component in one or more safety functions.
+
+The Linux operating system is not an item. Therefore, it makes no sense to say Linux or any software is safe. Linux is a general-purpose operating system, it is not designed for any specific use case. For special use cases, Linux can be customized and configured both at compile time and at runtime.
+
+As a component of a specific safety function, the instance of Linux that is individually configured and optimized for that use case must be functionally safe so that the safety function can reduce the danger posed by an item to an acceptable level. As a rule, specific real-time requirements are placed on the operating system for safety functions, among other things. For the mass of business-critical applications, for which Linux has proven itself so well in the past decades, such requirements are not made. This consideration alone gives a first indication that an analogy of applicability from mission-critical to safety-critical applications is not so simple.
+
+As a component of a safety function, Linux becomes part of the safety architecture of a particular item. For the risk analysis of the item, the risk that the safety function does not intervene in time, does not intervene at all or intervenes incorrectly in some way must then also be systematically analyzed and evaluated for each safety function.
+
+The systematic analysis and evaluation of risks is, at its core, a job that involves a lot of statistics and probability calculation. Probability theory is one of those mathematical disciplines where intuition and "common sense" often fail. This is the simple answer to the question of why Linux is not so easily a component for safety functions, even though it has proven itself so well for mission-critical applications. The simple analogy from mission-critical to safety-critical is intuitively obvious, but mathematically incorrect.
+
+A more accurate answer requires somewhat deeper considerations.
+The risk analysis for a technical system and its safety architecture is necessarily application-specific. Linux itself is only part of the safety function. In the specific mathematical consideration, a possible error of Linux in the safety function enters into a probability calculation. It is recognized and admitted that components can be used in a safety function which have proven themselves in other (as similar as possible) applications. Realistically, however, it must not simply be assumed that such a component is absolutely safe, i.e. that it functions correctly with probability 1.
+
+An exact value for the intuitively obvious reliability of Linux is difficult to determine because of the complexity of the monolithic kernel. In order to arrive at a reasonably accurate value for the purpose with a defined small uncertainty, a lot of testing has to be done, and probability calculation is again used.
+
+The mathematical details of the analysis established in IEC 61508 for proven in use are very well described, for example, at Exida and shall not be repeated in detail here. This way is possible, but unfortunately no longer intuitive. In IEC 61508 probabilistic limits are set for the SIL, which must be met for approval. For SIL 4, it is required that a dangerous failure occurs for a safety function with a high or continuous demand rate with an average frequency of less than 10-⁸/h (Probability of Dangerous Failure per Hour, PFHD). This means that proof must be provided that this safety function fails on average only every 11416 years. The limits from IEC 61508 for the PFHD have also been adopted in EN 50129 for the Tolerable Hazard Rate (THR). In ISO 26262-8 14.4.5.2.4, even 10-⁹/h is required for ASIL-D as the limit for the average frequency of actually observable malfunctions. For ASIL-B and ASIL-C, the limit of 10-⁸/h is set without difference.
+
+This is a somewhat more detailed answer to the question of why Linux is not so easily suitable as a component for safety functions, even though it has proven itself so well for mission-critical applications.
+
+The engineering consideration of functional safety has a history that goes back well beyond the existence of IEC 61508, ISO 26262 and EN 50128. Because the standards and norms that existed prior to 2001 did not adequately address the specifics of complex programmable electronic systems, IEC 61508 was written and published as a basic safety standard with a perspective strength for plant and process engineering. Following the publication of IEC 61508, the automotive industry developed its own standard tailored to its specific circumstances and requirements, which was first published in 2009 as ISO 26262. EN 50128 on software for railroad control and monitoring systems was first published in 2001.
+
+Passive and active electronic safety systems for road vehicles were developed and launched long before ISO 26262. Airbag (1981), ABS (1978) and ESP (1995) have considerably increased the safety of automobiles and significantly contributed to the prevention of traffic fatalities even without the standard requirements of ISO 26262.
+
+In order to be able to use such existing safety systems with the current standards, ISO 26262 in particular provides relatively detailed instructions compared with IEC 61508 and EN 50128 on how such safety systems, which have proven themselves in practice, can be approved under the new standard. In this context, the above-mentioned limits for the average frequency of actually observable malfunctions (10-⁹/h for ASIL-D, 10-⁸/h for ASIL-B and ASIL-C) are set as a prerequisite for compliance with the standard.
+
+A derivation leading to all the above limits (PFHD, THR, ...) is not provided in all the documents. The only reference is that the risk posed by an item must be reduced to an acceptable level. How this acceptable level is to be determined is not further explained. The concrete figures apparently reflect a value held at discretion for the assumed social consensus. From a societal point of view, there is nothing to be said against high safety measures for dangerous items. However, the concrete figures also have an (economic) political significance. ISO-26262 is not legally required for the registration of road vehicles (in Germany). Vehicles that do not comply with this standard can be imported and put into operation quite legally. This means that risk-taking or unscrupulous entrepreneurs, as well as manufacturers with the backing of sufficiently determined governments, are free to put vehicles without the high safety standards into operation even in Germany (Europe, the USA). With regard to the approval of Linux as an operating system for safety functions, it may be appropriate to question the established limits and to discuss realistic requirements politically as well.
+
+Compared to nuclear power plants or chemical plants, for example, unit numbers for safety systems in the automotive industry are comparatively high. In addition, the automobile manufacturers have a comprehensive and reliable database for their vehicles with fixed inspection intervals via the contract workshops, on which a mathematical proof of operational reliability can be provided. Therefore, it is not completely impossible for the automotive industry to provide this proof. However, this requires field data to be collected and evaluated over several years.
+
+> ToDo: Insert details about the PiU procedure in the automotive industry
+
+For open source software in general and Linux in particular, there is a database in the respective bug tracking systems, which can partly be seen as analogous to the workshop data of the automobile manufacturers. From these sources, however, the number of bug-free running systems and their uptime cannot be determined. Nor do they systematically record the details of the overall configuration of each individual system. However, all of these data are needed for the statistical analysis of functional safety.
+
+In order to underpin the advantages of Linux with the large open source communities for a proof of operational reliability in the context of functional safety with the necessary field data, a sufficiently large number of users would have to be prepared to regularly supply "inspection data" of their Linux servers to an analysis system, for example at the Linux Foundation. On the basis of such data, it could then be possible in principle to provide the proven in use argument for certain components and subsystems of the Linux operating system.
+
+
+Because Linux cannot be certified generally but always only for a concrete set of defined safety functions in each case, it is meaningful and necessary to conduct a proven in use argument for the configuration (compile and runtime) corresponding in each case to the application.
+
+Linux is not a microkernel, but it is very largely modular. For the purpose of a PiU argument, the statistically considered configuration can possibly be reduced to an absolute minimum.
+
+> ToDo: Consideration for decomposition and freedom from interference
+
+In practice as well as for risk assessment, device drivers may be considered particularly problematic compared to the scheduler or mature file systems. These components should therefore perhaps be excluded from the PiU argument. The drivers required for a specific safety function must then be designed, developed and tested in accordance with the standards. 
+
+In summary, it is not easy to derive suitability for safety-critical applications from the reliability of Linux as a server operating system. For such special use cases, high demands are placed on the evidence for error-free uptime. With the currently publicly available data, this proof cannot be provided with the currently required operating times. As a cooperative project of parties interested in functional safety for Linux, such an approach could perhaps be pursued further.
+
+In any case, it is reasonable and necessary for some components of the Linux kernel to follow the path described in the norms and standards for developing functionally safe software. Overall, it may also be beneficial for other use cases to develop software according to functional safety criteria. The associated requirements for analysis, design, documentation, testing and management are costly. Because this is open source development, parties interested in functional safety for Linux can work cooperatively here as well.
+
+ELISA provides a good framework for this collaboration.
+
+
+## References and further information
+
+* [Exida Determining Software Safety](https://www.exida.com/Resources/Whitepapers/determining-software-safety-understanding-the-possibilities-and-limitations)
+* [Math behind PiU](https://ez.analog.com/ez-blogs/b/engineerzone-spotlight/posts/the-math-behind-proven-in-use)
+* [Schlummer Beitrag zur Entwicklung einer alternativen Vorgehensweise für eine Proven-in-Use-Argumentation in der Automobilindustrie](https://d-nb.info/1022900226/34)
+


igor-stoppa (changes requested):
The problem I see with a proven in use argumentation is that it requires:

1. defining a very specific HW/SW combination
    *.  if you want to generalise the HW/SW combination with more pairs, you have to prove that they will be equivalent, with regard to the use for which you have historical data
2. you need to have evidence from operating that (class of) HW/SW combination in the field (meaning that it's not sufficient to say that it has worked, you need to have stats about MTBF, for example)
3. you need to prove that the HW/SW combination for which you want to make the proven in use argumentation is equivalent to the one for which you have gathered data
4. you need to prove that the use case and operating conditions to which this new HW/SW combination will be exposed are equivalent to those for which you have historical data.


If you can do THAT, there is no need to do much in terms of safety analysis and study of freedom from interference, because you have already proven that the new system is equivalent to one that would work, as long as they are used in the same scenario.

The fact that Linux is modular even if it is not a microkernel doesn't mean much - what the microkernel brings to the table is the HW-backed isolation.

What I was - and still am - VERY skeptic about are point #3 and #4, because (IMHO) it is very unlikely that you will find sufficient data about sufficiently homogenous systems used in a very similar way.

The fact that linux evolves so rapidly already makes it unrealistic to expect that one would have sufficient data for something sufficiently close (unless you want to use a 4.0 kernel, or example)

The point where linux not being a microkernel comes into play is that as soon as you need a new/different device driver, you have voided the assumption of similarity between your target system and the one which has produced the historical data.

The reason for being so strict in rejecting changes compared to the initial configuration is that, even assuming one might be able to re-use the very same HW and kernel, changing only user-space, that change alone might expose latent safety issues that have been there all the time - even while gathering data about how safe the system is - but never got a chance to be exposed.

By radically altering even the pattern by which memory is allocated, CPU is used, etc. the system ends up exerting very different activation patterns that are now outside the scope of the historical data, which becomes useless.

All in all, the arguments made in the pull request are unfit for a generic approach to "Proven in Use", because they do not really capture its purpose, which is all but generic.

I would rather suggest to pursue the use of PAS, because it's a much better fit and can actually yield some usable outcome.

reiterative:
@igor-stoppa wrote
> I would rather suggest to pursue the use of PAS, because it's a much better fit and can actually yield some usable outcome.

What does 'PAS' refer to in this context?

Proven-In-Use-Argument.md
```diff
@@ -0,0 +1,77 @@
+# Is Linux mature and proven enough for use in safety-critical applications?
+
+Linux has proven itself thousands, probably millions of times over in business-critical applications. So it makes sense to use Linux for safety-critical applications as well. Linux has many advantages
```
reiterative, line 3:
```suggestion
Linux has proven itself thousands, probably millions of times over in business-critical applications, so it may seem logical that we could use it for safety-critical applications as well. Linux has many advantages 
```

Proven-In-Use-Argument.md
```diff
@@ -0,0 +1,77 @@
...
+* it runs on an enormously wide range of hardware
+* there are a large number of established vendors and service providers for the operating system and for additional (standard) software
+* many specialists are familiar with it
+* it is continuously being developed
+* a large part of the innovation in IT is created on Linux and as open source software
+* ...
+
+Linux has obviously proven itself. So what is so difficult about safety-critical applications?
+
+First, it's useful to look briefly at what functional safety is actually about. At its core, it is about dangerous items: technical systems, devices, machines, vehicles and the like. Such items can harm the environment or people, and in the worst case can even kill people uninvolved in their operation.
```
reiterative, line 14: I think we should consider using 'product' or 'products' instead of 'item' or 'items' in this and other sentences. The word 'item' has a specific meaning in ISO 26262, which may cause readers to think that we are using the word in that sense in some circumstances. 

I would also make the following change:

```suggestion
First, it's useful to look briefly at what functional safety is actually about. At its core, it is about potentially dangerous items: technical systems, devices, machines, vehicles and the like, which can harm the environment or people, and in the worst case can even kill people uninvolved in their operation.
```

Proven-In-Use-Argument.md
igor-stoppa, line 3: There are several misconceptions already at the beginning:

1. "Linux", as intended in the document, doesn't exist. There are a large number of versions and revisions, each with own set of features and bugs.
2. When making claims of proven in use, one MUST refer to a specific SW configuration, its features and its bugs.
3. When making claims about proven in use, one MUST refer to a specific HW configuration.

There would be even the aspect of the toolchain, but the previous points are already sufficient to void the rest of the claims.

As I already stated, these sort of statements are not per se comepltely wrong, but they do nto belong to "Proven in Use", they rather fit in the PAS 8926 https://www.iso.org/standard/83346.html approach.

igor-stoppa:
> @igor-stoppa wrote
> 
> > I would rather suggest to pursue the use of PAS, because it's a much better fit and can actually yield some usable outcome.
> 
> What does 'PAS' refer to in this context?

The PAS 8926 https://www.iso.org/standard/83346.html

shetze:
> 1. defining a very specific HW/SW combination
>    *.  if you want to generalise the HW/SW combination with more pairs, you have to prove that they will be equivalent, with regard to the use for which you have historical data

Thank you for the comment, these are good objections.
If I understand correctly, the PiU is dealt with in ISO 26262-8. In Objective 14.1, the scope is "existing items or elements". From this I do not conclude immediately that elements are fixed HW/SW combinations. It would be good if we could reach a common understanding on this.

shetze:
> The fact that linux evolves so rapidly already makes it unrealistic to expect that one would have sufficient data for something sufficiently close (unless you want to use a 4.0 kernel, or example)

I totally agree that the PiU argument is not easy to make. However, my understanding is that it has certainly been applied in the automotive industry. Here it would be good if someone could contribute concrete data on how often this has actually been the case so far.

However, with the limits given in the current standard (10-⁹ for ASIL-D), it is not easy to prove the required operating times even for the automotive industry with the large numbers of units compared to other technical systems. If I understand correctly, the automotive industry expects an average of 400 operating hours per year. The series are quite respectable with perhaps 10k to 20k, but still small in view of 11416 years.

For Linux servers, however, we can count on a number of operating hours that is larger by a factor of 20. For a constructive handling of the challenge, I suggest to formulate the concrete requirements for the data to be collected. We can then leave the question of whether it is realistically possible to collect this data to the stakeholders interested in such proof.

Proven-In-Use-Argument.md
shetze: I see the problem with the term item although it does intentionally lead into the direction of its use in ISO 26262. Product appears to soft for me. What do you think about thing?

igor-stoppa:
> > 1. defining a very specific HW/SW combination
> >    *.  if you want to generalise the HW/SW combination with more pairs, you have to prove that they will be equivalent, with regard to the use for which you have historical data
> 
> Thank you for the comment, these are good objections. If I understand correctly, the PiU is dealt with in ISO 26262-8. 

In Objective 14.1, the scope is "existing items or elements". From this I do not conclude immediately that elements are fixed HW/SW combinations. It would be good if we could reach a common understanding on this.

The fact that there needs to be a fixed HW/SW combination stems from the need to have homogeneous data.
Within reason one can argue (and prove) that several HW/SW combinations are equivalent (think of a node shrink in the processor), but lacking this uniformity across units which have produced the historical data, you cannot claim tha the data is homogensous either.

The ISO desn't get into details because it's up to specific HW/SW instances how much the can be generalised.
With a microkernel, for exampl,e you could claim that device drivers are sufficiently isolated to not be able to matter much, and therefore you would get a better shot at grouping together - for example - variants of a same device, because their core would be more consistent. But with Linux you do not have such isolation, and hterefore the same claim would not hold.

You cannot expect the ISO to give direct guidance on specific applications, because it would most likely miss many sub-cases. Instead, it gives generic guidelines and it is up to who deals with a certain application to understand how the ISO applies.

igor-stoppa:
> > The fact that linux evolves so rapidly already makes it unrealistic to expect that one would have sufficient data for something sufficiently close (unless you want to use a 4.0 kernel, or example)
> 
> I totally agree that the PiU argument is not easy to make. However, my understanding is that it has certainly been applied in the automotive industry. Here it would be good if someone could contribute concrete data on how often this has actually been the case so far.
> 
> However, with the limits given in the current standard (10-⁹ for ASIL-D), it is not easy to prove the required operating times even for the automotive industry with the large numbers of units compared to other technical systems. If I understand correctly, the automotive industry expects an average of 400 operating hours per year. The series are quite respectable with perhaps 10k to 20k, but still small in view of 11416 years.
> 
> For Linux servers, however, we can count on a number of operating hours that is larger by a factor of 20. For a constructive handling of the challenge, I suggest to formulate the concrete requirements for the data to be collected. We can then leave the question of whether it is realistically possible to collect this data to the stakeholders interested in such proof.

The concrete requirement, as I stated earlier, and it doesn't come from me, but from the ISO and the way it is applied (please check it with Gab, for example) is that you need to have a high level of uniformity between HS/SW that has produced the data, the one you want to use now, the use cases that have produced the data and the use case you intend to use.

Unless you plan to put a server in acar and use similar sw, I see little use for that data colelcted from servers.

I do not how else to say it: it's clear you really wan to use linux, and so we all do, but if it was just a matter of making these claims you are attempting to do, don't you think that it would have been already done a long time ago?

Proven-In-Use-Argument.md
reiterative: Upon further consideration, I'm happy with item

Proven-In-Use-Argument.md
shetze: I still do not accept to give up the option of PiU alltogether. Can you agree with the following text? Would this make the problem sufficiently clear and still provide a constructive path for the argument?

"""
It is obvious and must nevertheless be clearly repeated: Linux can never be approved 'as a whole' and 'in general' for use in safety-critical applications. There are any number of examples of Linux configurations and constellations in which the operating system fails mercilessly. Therefore, according to the conservative interpretation of the standard documents, only individual, fully specified systems can prove themselves. In this sense, every Linux system is unique and a PiU argument is categorically ruled out. 

In order to approach the PiU argumentation for Linux, the dimension of the system under consideration or the elements under consideration must be reduced and focused on the one hand. On the other hand, value ranges for configuration and constellation must be described that are recognized as equivalent with regard to the safety-critical application. If it is possible to document the elements in focus in running systems with their behavior, especially in the event of system errors and in extreme situations, and to document robust behavior, a PiU argumentation could be based on this.
"""

Proven-In-Use-Argument.md
```diff
@@ -0,0 +1,77 @@
...
+
+For obvious reasons, the dangers posed by an item must be reduced to an acceptable level before the item is put into operation or placed on the market. 
+Thus, to reduce the unavoidable risks from an item to an acceptable level, appropriate safeguards must be found and implemented at the design stage. These protective measures come into action under certain conditions and regulate the item in such a way that a previously identified and determined risk remains below a predetermined threshold. These protective measures are called safety functions. In certain situations, a safety function can put the item in a state recognized as safe, for example, by shutting the item down.
+
+With the introduction of safety functions, the risk posed by an item now depends on the correct operation of that protective measure. Thus, the item is safe only to the extent that the safety function operates correctly and reliably. This is the safety of the safety function, or simply functional safety.
```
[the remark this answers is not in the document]
shetze: This formulation is perhaps a little abbreviated, but I'm actually quite proud of it. Do you think it's wrong?

In Wikipedia I find the following for [functional safety](https://en.wikipedia.org/wiki/Functional_safety)

`Functional safety is the part of the overall safety of a system or piece of equipment that depends on automatic protection operating correctly in response to its inputs or failure in a predictable manner (fail-safe).`

Somewhat abbreviated that means:

`Functional safety is the part of the overall safety of a system that depends on automatic protection operating correctly in a predictable manner.`

The automatic protection implemented in software is the safety function, this leads to safety of the safety function. Can you go along with that?

Proven-In-Use-Argument.md
igor-stoppa: In my opinion, the initial problem lies in the wording "Proven in Use" - it is already widely asociated to certain concepts which do not match what you are after. I do not understand why you seem unmovable on this.
Of course you can arbitrarily redefine words and expressions, to mean what you wish them to mean.
However, it seems an unnecessary uphill battle.

Then, there are a bunch of further considerations on several assumptions that you seem to make, like what is the "installed base" that can be taken as reference, what is the accept/reject criteria.

For example: a handheld console, a water plant management system, a server,  a drone, can all run linux.
How do you define which ones can be taken as reference and which cannot?
Assuming that you have defined such a set, which changes can one make and still keep the claims valid and which would void the claims.

To me, this whole argumentation looks very verbose, but it lacks the clarity of scope that I would expect when making such claims.

It is a fact that - usually - one will want to take something as baseline and alter it.

This whole thing is missing clear criteria for what is admissible, what is not, and what is admissible, but introduces extra validation (for example).

I'd rather see a set of conicise bullet points with ok/nok do/don't and so on.

Proven-In-Use-Argument.md
```diff
@@ -0,0 +1,77 @@
...
+
+It is obvious to implement safety functions at least partially with electronic or programmable components. It is only in this context that the question of using Linux for safety-critical applications arises: as an operating system for a programmable component in one or more safety functions.
+
+The Linux operating system is not an item. Therefore, it makes no sense to say Linux or any software is safe. Linux is a general-purpose operating system, it is not designed for any specific use case. For special use cases, Linux can be customized and configured both at compile time and at runtime.
```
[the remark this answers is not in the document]
igor-stoppa: I'd argue that nothing here is very obvious, since the discussion on htese topics has been going on for years.
If you want to make statements, provide backing evidence.

shetze:
My aim is still to produce a positive, constructive text on the subject of PiU. This is important to me because ELISA in particular can also aspire to criticize the existing formulations and established interpretations of the current standards and suggest changes. Igor refers to PAS 8926 in which this has been done for ISO 26262 and the qualification of software. Why shouldn't it be possible for IEC 61508 and PiU, for example?

I would also like to point out that, as far as I can see, the argument against PiU assumes that there will be no SIL-2 safety function with Linux kernel support. That is fatalistic.
If we assume that Red Hat is successful with RHIVOS and the system is used by GM in series production vehicles, isn't that exactly the context for a PiU argument? Why shouldn't we at least formulate what information GM would have to collect and publish in order to base a PiU argument on it?

And why shouldn't we demand that such a PiU argument is also valid across sector boundaries? Automated Train is precisely about making the transfer from the automotive industry to the rail sector. That's why I'm interested in PiU.

If I understand correctly, such calls to improve the standards have already been made. We should reference a few of these attempts from the past.

https://gitlab.opentech.at/markus/sil2linuxmp/-/blob/main/doc/literature/HSE/crr01337.pdf
https://www.simula.no/research/projects/opencoss-open-platform-evolutionary-certification-safety-critical-systems/
https://d-nb.info/1022900226/34

igor-stoppa:
> My aim is still to produce a positive, constructive text on the subject of PiU. This is important to me because ELISA in particular can also aspire to criticize the existing formulations and established interpretations of the current standards and suggest changes. Igor refers to PAS 8926 in which this has been done for ISO 26262 and the qualification of software. Why shouldn't it be possible for IEC 61508 and PiU, for example?
> 
> I would also like to point out that, as far as I can see, the argument against PiU assumes that there will be no SIL-2 safety function with Linux kernel support. That is fatalistic. If we assume that Red Hat is successful with RHIVOS and the system is used by GM in series production vehicles, isn't that exactly the context for a PiU argument? Why shouldn't we at least formulate what information GM would have to collect and publish in order to base a PiU argument on it?
> 
> And why shouldn't we demand that such a PiU argument is also valid across sector boundaries? Automated Train is precisely about making the transfer from the automotive industry to the rail sector. That's why I'm interested in PiU.
> 
> If I understand correctly, such calls to improve the standards have already been made. We should reference a few of these attempts from the past.
> 
> https://gitlab.opentech.at/markus/sil2linuxmp/-/blob/main/doc/literature/HSE/crr01337.pdf https://www.simula.no/research/projects/opencoss-open-platform-evolutionary-certification-safety-critical-systems/ https://d-nb.info/1022900226/34

Let's forget for a moment all the standards, ISO or not.
Let's instead look at it in layman's terms.
"Proven in Use" means:
 "This thing has perfromed well. I do not even need to knwo what is inside, there could be a chipmunk running on a wheel, for all I know; or a little gnome. But it works well enough, for the purposes I have used it so far. I know it, because I have record which prove it. If I re-use it, as-it-is, for the same purpose, I believe that it will work equally well."
 
 But, as soon as you want to change something, either in the use case, or in the components - let's say swap the chipmunk for an hamster or the gnome for an elf, THEN you have to prove that they are REALLY equivalent."
 
 This is the sort of reasoning that I do not see in all the stuff you wrote.
 Do you want to put a GM vehicle with RHIVOS on a railroad track? Something else? What is the argumentation that they are equivalent? What is the argumentation that the data collected by driving cars is good enough for trains?
 Do you want to use exactly the same hardware and software? Do you want to change anything in them? What? What is the criteria for accepting/rejecting certain changes? Are all changes ok?
 
 Personally, I am not interested in discussing only the astronaut-architect perspective, where very practical problems are not considered.

Proven-In-Use-Argument.md
```diff
@@ -0,0 +1,76 @@
...
+
+In order to underpin the advantages of Linux with the large open source communities for a proof of operational reliability in the context of functional safety with the necessary field data, a sufficiently large number of users would have to be prepared to regularly supply "inspection data" of their Linux servers to an analysis system, for example at the Linux Foundation. On the basis of such data, it could then be possible in principle to provide the proven in use argument for certain components and subsystems of the Linux operating system.
+
+Because Linux cannot be certified generally but always only for a concrete set of defined safety functions in each case, it is meaningful and necessary to conduct a proven in use argument for the configuration (compile and runtime) corresponding in each case to the application.
+
+It is clear from the explanations so far that a PiU argument is already associated with a great deal of effort for application within a domain. However, the final exclusion criterion for the use of PiU comes from the practice of approval procedures. PiU is generally viewed very skeptically in the safety community. ISO 26262 Part 8 Clause 14 is not an option.
```
reiterative, line 58:
```suggestion
It is clear from the explanations so far that a PiU argument is already associated with a great deal of effort for application within a domain. However, the final exclusion criterion for the use of PiU comes from the practice of approval procedures. PiU is generally viewed very skeptically in the safety community. ISO 26262 Part 8 Clause 14, which details how a PiU argument may be applied, makes it clear that this is not an option for Linux, because it only applies to a "type of product whose definition and conditions of use are identical to or have a very high degree of commonality with a product that is already released and in operation". 
```

Proven-In-Use-Argument.md
```diff
@@ -0,0 +1,76 @@
...
+
+For the automotive sector, software developed outside of a safety context can be qualified in accordance with ISO 26262 Part 8 Clause 12. Outside this domain, there is no provision for cross-domain qualification. The continuous repair of vulnerabilities and errors, which is a matter of course for Linux, is also unthinkable within the established certification practice.
+
+The specific approaches to certifying Linux for the automotive sector do not use the PiU argument. Whether future changes in the standard texts and in certification practice will make a provenance argument and the transfer of safetey elements from one domain to another possible is the subject of science fiction.
```
reiterative, line 62:
```suggestion
The specific approaches to certifying Linux for the automotive sector cannot be based on a PiU argument. Whether future changes in the standard texts and in certification practice will make a provenance argument and the transfer of safety elements from one domain to another possible is the subject of science fiction.
```

Proven-In-Use-Argument.md
```diff
@@ -0,0 +1,76 @@
...
+
+Compared to nuclear power plants or chemical plants, for example, unit numbers for safety systems in the automotive industry are comparatively high. In addition, the automobile manufacturers have a comprehensive and reliable database for their vehicles with fixed inspection intervals via the contract workshops, on which a mathematical proof of operational reliability can be provided. Therefore, it is not completely impossible for the automotive industry to provide this proof. However, this requires field data to be collected and evaluated over several years.
+
+> ToDo: Insert details about the PiU procedure in the automotive industry
```
reiterative, line 50:
```suggestion
```
I don't believe that additional material would add anything here.

Proven-In-Use-Argument.md
```diff
@@ -0,0 +1,76 @@
...
+
+In order to underpin the advantages of Linux with the large open source communities for a proof of operational reliability in the context of functional safety with the necessary field data, a sufficiently large number of users would have to be prepared to regularly supply "inspection data" of their Linux servers to an analysis system, for example at the Linux Foundation. On the basis of such data, it could then be possible in principle to provide the proven in use argument for certain components and subsystems of the Linux operating system.
+
+Because Linux cannot be certified generally but always only for a concrete set of defined safety functions in each case, it is meaningful and necessary to conduct a proven in use argument for the configuration (compile and runtime) corresponding in each case to the application.
```
reiterative, line 56:
```suggestion
Because Linux cannot be certified generally but always only for a concrete set of defined safety functions in each case, it is meaningful and necessary to conduct a proven in use argument for the configuration (compile and runtime) corresponding in each case to the application. Thus, in order to use this kind of data to support a safety case, the data would need to have been collected for a sufficiently well-defined deployment and configuration of Linux in an relevant context (e.g. as part of vehicle component, for an automotive safety case), with a comparable role with respect to the functional safety of that item.
```

reiterative:
Overall, I think this is a useful discussion of the topic, which explains why an apparently intuitive 'proven in use' argument cannot be applied, and explores both how such an argument is applied to other kinds of system,  as well as some of the alternative strategies for arguing that pre-existing software is acceptably safe in a given context.

I have made some suggestions, and Lukas has requested a few more tweaks. However, I think we are not too far from having a document that we can publish. Can we resolve some of the other threads e.g. those from @igor-stoppa by adding a note in the text to make it clear that a claim is disputed, or including an alternative framing alongside it?

Proven-In-Use-Argument.md
```diff
@@ -0,0 +1,77 @@
...
+
+As a component of a specific safety function, the instance of Linux that is individually configured and optimized for that use case must be functionally safe so that the safety function can reduce the danger posed by an item to an acceptable level. As a rule, specific real-time requirements are placed on the operating system for safety functions, among other things. For the mass of business-critical applications, for which Linux has proven itself so well in the past decades, such requirements are not made. This consideration alone gives a first indication that an analogy of applicability from mission-critical to safety-critical applications is not so simple.
+
+As a component of a safety function, Linux becomes part of the safety architecture of a particular item. For the risk analysis of the item, the risk that the safety function does not intervene in time, does not intervene at all or intervenes incorrectly in some way must then also be systematically analyzed and evaluated for each safety function.
```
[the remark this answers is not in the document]
reiterative: I will try to clarify this sentence

Proven-In-Use-Argument.md
```diff
@@ -0,0 +1,74 @@
...
```
reiterative, line 27:
```suggestion
As a component of a safety function, Linux becomes part of the safety architecture of a particular item. For the risk analysis of the item, the risk that a component of the OS that is involved in implementing the safety function does not fulfil its responsibility (e.g. does not intervene in time, or does not intervene at all or intervenes incorrectly in some way) must then also be systematically analyzed and evaluated for each safety function.
```

Proven-In-Use-Argument.md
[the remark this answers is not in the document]
reiterative: See https://github.com/elisa-tech/wg-osep/pull/21/files#r1419077886

Proven-In-Use-Argument.md
```diff
@@ -0,0 +1,77 @@
...
+
+The Linux operating system is not an item. Therefore, it makes no sense to say Linux or any software is safe. Linux is a general-purpose operating system, it is not designed for any specific use case. For special use cases, Linux can be customized and configured both at compile time and at runtime.
+
+As a component of a specific safety function, the instance of Linux that is individually configured and optimized for that use case must be functionally safe so that the safety function can reduce the danger posed by an item to an acceptable level. As a rule, specific real-time requirements are placed on the operating system for safety functions, among other things. For the mass of business-critical applications, for which Linux has proven itself so well in the past decades, such requirements are not made. This consideration alone gives a first indication that an analogy of applicability from mission-critical to safety-critical applications is not so simple.
```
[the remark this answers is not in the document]
reiterative: @bulwahn What would address your concerns here? Should we omit the reference to real-time requirements, for example? Could you suggest a rewording?

PR #36, opened 2024-05-02
Example of system decomposition

igor-stoppa:


Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,167 @@
...
+
+## **Structure of the document**
+The document is divided in two parts, based on hte destinatary of the memory allocations discussed: kernel-space or user-space.
+Individual points are numbered for ease of reference, but the numbering is not meant to represent any sequence.
+
+## **Memory management in Linux**
+
+### **Kernel-space memory allocations**
+
+#### **Facts**
```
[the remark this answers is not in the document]
reiterative: Yes, I suggest that you remove the 'Fact' sub-headings and add a preamble stating that these sections list *assertions* about Linux behaviour that "can be verified by inspection of the relevant code" or words to that effect.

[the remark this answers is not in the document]
igor-stoppa: I did not particularly like the "Facts" part either, what I was/am trying too do is to separate "objectively verifiable self-contained truths" - as in "it's in the code" - from parts that are inferred. I'll follow @reiterative 's proposal

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,167 @@
...
+1. unlike processes memory, kernel memory pages are not swapped, nor dropped silently by the kernel itself,
+   although an hypervisor will do to a VM what the kernel does to a process (but this is beyond the control of the kernel)
+2. the kernel context (usually EL1 on ARM64) uses one single memory map (page tables) across all the cores
+   executing in kernel mode
+3. on 64 bit systems (e.g. ARM64 and x86_64), usually almost all physical memory is mapped in a
+   (semi)contiguous (there can be holes) range. Memory within this range is both virtually and physically contiguous.
+4. physically contiguous memory is treated as a scarce resource, and typically is not provided to userspace,
+   unless it explicitly asks for it (e.g. for implementing some DMA buffer)
+5. the kernel can access userspace in two ways:
```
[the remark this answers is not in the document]
igor-stoppa: ACK

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,167 @@
...
+   1. through the userspace mapping
+      1. it is done in exceptional cases, like copy_to_user_copy_from_user
```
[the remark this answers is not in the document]
igor-stoppa: @dweingaertner : "exceptional cases" was referred to the fact that - codewise (spatially?) - the kernel does not have acces to the memory map of a process, but now I see that it might be misunderstood in a temporal sense, where the kernel indeed accesses  userspace memory all the time, just through a limited set of code paths.
I will have to be more verbose on this.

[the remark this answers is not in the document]
igor-stoppa: @shetze yes, you are right, I forgot those 2 - I will add them, thanks. WRT the meaning of "exceptional", please read my comments above. But yes, it's obviously not well worded and I will try to describe it better.

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,167 @@
...
+      2. it uses the userspace memory map, which can implement HW protections against kernel misuses
+      3. it allows the kernel to see process pages in the same sequence and with the same mappings as the proces does
+   2. through the EL1 mappings + linear map
```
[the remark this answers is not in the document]
igor-stoppa: * ok on the style 
* the security aspect is there because I'm afraid that if I do not explain it upfront, it will be periodically brought up, since I noticed that people tend to confuse the intent behind certain security features and think that said features provide more protection that they actually do
* the kernel doesn't intrinsically "know" if a page is in use by user space or by itself, because in both cases it is always mapped. The problem is that a broken driver or a broken kernel component can cause cascaded effects, so it's not so straightforward to say where the wrong access cancome from. It might come from a legitimate and "sane" component that has had its metadata corrupted by another one. I can elaborate more on this.

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,167 @@
...
+      1. not the intended way to access process context
+      2. not very useful, even for an attacker, in a security scenario
+      3. the sequence of pages mapped in the userspace process is not known
+      4. some process pages might even be "missing", because they have been either swapped out or dropped
+      5. bypasses any protection that the process might employ through its own mapping
+6. attempt to keep pages hot in the system cache make it very difficult to predict which pages might surface in a certain place:
```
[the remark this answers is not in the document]
igor-stoppa: * suface -> emerge, in the sense that one can look at the underlying memory management features and htnk that a certain page is acquired, released, and then it gets recycled elsewhere.
* the intent is to show how it is practically impossible to make any realistic assessment about the health of the memory subsystem, because of its intricacy and its exposure to a very large amount of fleeting perturbations, from both qualified and qm components, both from the kernel itself and user-space, driven by both internal and external events.

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,167 @@
...
+   4. However, it also attempts to cache locally these free pages, in per-cpu queues, with the intent of minimising
+      the amount of cache flushes that would be driven by releasing a recently feed page, and subsequently allocating
+      another one
+
+
+### **User-space memory allocations**
+
+#### **Facts**
+1. user-space memory must be assumed to not always be backed by physical memory pages
+   meaning that a non running process might have had some of its pages dropped or swapped out to disk
```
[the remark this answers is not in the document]
igor-stoppa: True, but I was trying to formulate the minimum simple condition. Not sure of the value of the attempt, but that is what I ahd in mind when I wrote that sentence.

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,167 @@
...
+2. unless some extra effort has been taken, to pin down memory allocations for user-space, it is
```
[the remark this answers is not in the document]
igor-stoppa: It was kind of stating what you pointed out in your previous comment, that even a running process could have some pages dropped/swapped, unless it has taken some care to prevent it.

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,167 @@
...
+   2. file-backed mappings:
+      1. constants
+      2. the main executable associated with the process
+      3. the runtime linker (optional but typical)
+      4. linked libraries
+4. the kernel uses various optimisations for dealing with processes on-demand mapping:
+   1. a physical page is allocated and mapped to a process only when the process accesses it:
+      1. file backed pages are allocated/mapped only when read/written to
+      2. anonymous pages are allocated only when written to
+   2. zero page: when a page is known to be empty, it is not reserved and mapped; instead,
```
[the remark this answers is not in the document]
igor-stoppa: ACK

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,167 @@
...
+      the kernel has 1 specific memory page that is mapped read-only
+   3. shared libraries:
+      when the same library pages are mapped by multiple processes, the library physical pages
+      are mapped as read-only into each process address space
+   4. copy-on-write:
+      1. when a library has own data, this is initially mapped as read-only and shared;
+         only when written to, then a separate physical page is reserved
```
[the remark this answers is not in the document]
igor-stoppa: OK

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,167 @@
...
+      2. same happens for data pages that were treated as zero-page, but then are written to.
+   5. folios: data structures that try to better abstract compound pages and *might* also be used
+      to represent optimised contiguous pages on ARM64 (instead of mapping 16 entries, it is possible
+      to map a 16-pages chunk of physically contiguous memory, that is also aligned to a 16 pages
+      boundary, reducing TLB use).
+      1. A folio acts as intermediary between vma and lower level memory management
+      2. it might pre-allocate/map more pages than explicitly requested
```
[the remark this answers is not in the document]
igor-stoppa: I have spent too much time on these things and tend to take some parts for granted :-D 
you are right, it's not obvious. My intent was still to show how impossible it is to predict the behaviour of the kernel, based on external stimuli applied, because of the high degree of both autonomy and complexity it has, in trying to optimise operations.

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,167 @@
...
+5. read-ahead: when asked to fetch data from disk, the kernel might attempt to optimise the operation,
+   reading more pages than requested, under the assumption that more requests might be coming soon
```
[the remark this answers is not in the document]
igor-stoppa: We care because it is even more evidence of the unpredictability of hte kernel with regard to llocation of memory pages.
E.g. if one even attempted to reach a point where transients have completed and all memory had been allocated and put into use, it would not be trivial to verify that such a state has been reached, and to define the steps to reach it, even allowing for some extra ones "just in case".

Contributions/Linux_Memory_Management_Essentials.md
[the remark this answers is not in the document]
igor-stoppa: Done

Contributions/Linux_Memory_Management_Essentials.md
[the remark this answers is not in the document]
igor-stoppa: The rewording should now address the comments above.

dweingaertner (changes requested):
I'm not sure if you are still working on the text, but here are my suggestion after your last comments. Hope they help

Contributions/Linux_Memory_Management_Essentials.md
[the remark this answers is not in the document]
igor-stoppa: Fixed in next version

Contributions/Linux_Memory_Management_Essentials.md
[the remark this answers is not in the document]
igor-stoppa: done

Contributions/Linux_Memory_Management_Essentials.md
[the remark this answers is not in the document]
igor-stoppa: I put in the kernel rant some more information about folios

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -60,17 +62,19 @@ Individual points are numbered for ease of reference, but the numbering is not m
    (semi)contiguous (there can be holes) range. Memory within this range is both virtually and physically contiguous.
 4. physically contiguous memory is treated as a scarce resource, and typically is not provided to userspace,
    unless it explicitly asks for it (e.g. for implementing some DMA buffer)
-5. the kernel can access userspace in two ways:
+5. the kernel can access userspace memory in two ways:
    1. through the userspace mapping
-      1. this is done in exceptional cases, like copy_to_user_copy_from_user
-      2. it uses the userspace memory map, which can implement HW protections against kernel misuses
-      3. it allows the kernel to see process pages in the same sequence and with the same mappings as the proces does
+      1. this is type of access is limited to few functions, like copy_to_user()/copy_from_user() and put_user()/get_user()
```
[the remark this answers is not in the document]
igor-stoppa: yes, thanks, I had noticed it and removed it; i've also expanded it a bit, to make it more clear

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -60,17 +62,19 @@ Individual points are numbered for ease of reference, but the numbering is not m
...
+      2. outside the execution of the functiona mentioned, this type of access is not possible, because the userspace mappings are made available only while executing those functions
+      3. it uses the userspace memory map, which can implement HW protections against kernel misuses
+      4. it allows the kernel to see process pages in the same sequence and with the same mappings as the proces does
```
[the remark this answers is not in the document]
igor-stoppa: fixed, thanks

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -60,17 +62,19 @@ Individual points are numbered for ease of reference, but the numbering is not m
...
 4. physically contiguous memory is treated as a scarce resource, and typically is not provided to userspace,
    unless it explicitly asks for it (e.g. for implementing some DMA buffer)
-5. the kernel can access userspace in two ways:
+5. the kernel can access userspace memory in two ways:
    1. through the userspace mapping
-      1. this is done in exceptional cases, like copy_to_user_copy_from_user
-      2. it uses the userspace memory map, which can implement HW protections against kernel misuses
-      3. it allows the kernel to see process pages in the same sequence and with the same mappings as the proces does
+      1. this is type of access is limited to few functions, like copy_to_user()/copy_from_user() and put_user()/get_user()
+      2. outside the execution of the functiona mentioned, this type of access is not possible, because the userspace mappings are made available only while executing those functions
```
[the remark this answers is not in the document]
igor-stoppa: I think it should be fixed in the new version

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -60,17 +62,19 @@ Individual points are numbered for ease of reference, but the numbering is not m
...
+      3. it uses the userspace memory map, which can implement HW protections against kernel misuses
+      4. it allows the kernel to see process pages in the same sequence and with the same mappings as the proces does
    2. through the EL1 mappings + linear map
-      1. not the intended way to access process context
-      2. not very useful, even for an attacker, in a security scenario
-      3. the sequence of pages mapped in the userspace process is not known
-      4. some process pages might even be "missing", because they have been either swapped out or dropped
-      5. bypasses any protection that the process might employ through its own mapping
+      1. not the intended way to access process context, because the intended way is through userspace mappings
+      2. attempting to use EL1 mappings would be not very useful:
```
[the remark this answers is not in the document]
igor-stoppa: I'm really sorry I failed to notice these suggestions; I ended up rearranging anyways most of the text, based on previous comments you gave, hopefully at least in spirit I have followed your suggestions

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -60,17 +62,19 @@ Individual points are numbered for ease of reference, but the numbering is not m
...
+         1. even for an attacker, in a security scenario (this is why EL1 mappings are allowed to have acccess to pages containing EL0 code/data)
```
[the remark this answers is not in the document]
igor-stoppa: not applicable anymore, unfortunately

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -80,10 +84,16 @@ Individual points are numbered for ease of reference, but the numbering is not m
       the amount of cache flushes that would be driven by releasing a recently feed page, and subsequently allocating
```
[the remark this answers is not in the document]
igor-stoppa: I ended up rewriting this part

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,179 @@
...
+
+### **User-space memory allocations**
+
+#### **Assertions**
+The following section presents a set of statements that can be objectively verified e.g. inspecting the sources.
+
+1. user-space memory must be assumed to not always be backed by physical memory pages
+   meaning that a non running process might have had some of its pages dropped or swapped out to disk
+2. unless some extra effort has been taken, to pin down memory allocations for user-space, it is
+   not really possible to assume much, other than at some point both code that has been executed
```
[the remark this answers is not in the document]
igor-stoppa: victim of the rewrite

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,179 @@
...
+   and data that has been read were present in physical memory. The presence of caching and other
```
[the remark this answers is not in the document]
igor-stoppa: outdated

igor-stoppa:
@reiterative the PR title is updated to better reflect its content. As I wrote also in a mail the First PRinciples statments need the Memory Management Essentials to be fully justified, so I grouped them together.
I wrote to Pete asking if he agrees to be mntioned as co-author, but I have not yet added him, till I get permission - and his email address.

igor-stoppa:
Added Pete to the list of co-authors

Contributions/Linux_Kernel_Safety_First_Principles.md
```diff
@@ -0,0 +1,141 @@
...
+*  However, it is possible to describe the Linux Kernel from a safety point of view.
+*  Not all the issues listed will be necessarily relevant to each system/application.
+*  A Safety Analysis should, however, clearly state which issues are relevant, which are not, and why.
+
+
+## **First Principles - Integrity**
+This section provides touchstone concepts about interference of QM components with
+safety-qualified ones.
+
+1. The vanilla Linux Kernel, is not sufficient to support, alone, safey goals.
```
paolonig, line 70: IMO this statement is misleading as it can be read as a negative conclusion. IMO it should be rephrased as "To support safety goals, The vanilla Linux Kernel must be accompanied by appropriate evidences/argumentations"

Contributions/Linux_Kernel_Safety_First_Principles.md
```diff
@@ -0,0 +1,141 @@
...
+2. No internal kernel barriers/protections against interference to variable data, including safety-relevant data.
```
paolonig, line 71: What does this mean? variables stored in stack for example can be protected by CONFIG_STACKPROTECTOR to some degree. IMO we should say "the monolithic nature of the Kernel poses challenges in protecting safety-relevant data against interference"

Contributions/Linux_Kernel_Safety_First_Principles.md
```diff
@@ -0,0 +1,141 @@
...
+3. Any component within the kernel context can generate (cascaded) interference.
+4. Internal interference can cascade also through components that have been qualified for safety at unit-level.
```
paolonig, line 73: I would remove "at unit-level" as the concept of unit, usually, applies to newly developed SW, not when qualifying pre-existing one

Contributions/Linux_Kernel_Safety_First_Principles.md
```diff
@@ -0,0 +1,141 @@
...
+5. The Linux Kernel is able to interfere with both itself and any part of user-space processes (linear map).
+6. No generic solution for very complex systematic corruption of any writable memory.
```
paolonig, line 75: Here what do you mean by "generic solution"? Do you mean "currently there is no design solution to protect writable safety relevant memory"?

Contributions/Linux_Kernel_Safety_First_Principles.md
```diff
@@ -0,0 +1,141 @@
...
+7. Stress testing alone is not sufficient for supporting safety claims.
```
paolonig, line 76: This is conclusion that IMO is not appropriate. An OEM could stress test the OS in a car for as many hours as needed to claim proven in use for example. IMO it can be reprhased as "It is usually difficult to support a safety claim exclusively by stress testing"

Contributions/Linux_Kernel_Safety_First_Principles.md
```diff
@@ -0,0 +1,141 @@
...
+8. Safety claims on the Linux Kernel require both positive testing and focused negative testing.
```
paolonig, line 77: Also here the statement is too absolutist. I would rephrase as "For requirement based testing usually both positive testing as well as focused negative testing are needed"

Contributions/Linux_Kernel_Safety_First_Principles.md
```diff
@@ -0,0 +1,141 @@
...
+9. Those security features which are based on randomisation decrease repeatability of testing (e.g. structure layout randomisation).
+10. Safety claims must be supported by components with the same-or-better safety level (e.g. a safety-qualified watchdog).
```
paolonig, line 79: incomplete sentence IMO. I would say: Safety claims must be supported by components with the same-or-better safety level than the ones associated with them (e.g. a safety-qualified watchdog)

Contributions/Linux_Kernel_Safety_First_Principles.md
```diff
@@ -0,0 +1,141 @@
...
+11. Non safety qualified processes can interfere with safety qualified ones, indirectly, through the kernel (e.g. triggering memory management bugs).
```
paolonig, line 80: IMO this sentence expose the final (bad) effect without focusing on the topic. I would say: The Kernel plays a key role for supporting the claim of coexistence of processes of different criticality (e.g. a memory management bug could trigger interferences between such processes).

Contributions/Linux_Kernel_Safety_First_Principles.md
```diff
@@ -0,0 +1,141 @@
...
+12. Using cgroups/containers/SELinux removes only certain simple types of user-space-induced interference.
```
paolonig, line 81: I would remove "simple" as it is too much debatable IMO

Contributions/Linux_Kernel_Safety_First_Principles.md
```diff
@@ -0,0 +1,141 @@
...
+13. In a mixed-criticality scenario, non safety-qualified code represents a safety liability that grows with the frequency of execution (cgroups/containers, LSM/SELinux, etc.).
```
paolonig, line 82: Are you saying that the more often NSR code is executed the higher the chance of interference is? If so why are you mentioning cgroups/containers ?

Contributions/Linux_Kernel_Safety_First_Principles.md
```diff
@@ -0,0 +1,141 @@
...
+14. HW enhancements are not a catch-all solution (e.g. ECC Memory doesn't prevent interference from non-safety-qualified SW).
+15. Detection of interference depends more on the recipient than the origin of the interference: too many sources, too difficult to model reliably.
```
paolonig, line 84: I don't think it is possible to agree in general with such a statement. E.g. KASAN would detect interference at the origin, not at the recipient...

Contributions/Linux_Kernel_Safety_First_Principles.md
```diff
@@ -0,0 +1,141 @@
...
+16. Difficult to say when all necessary resources are allocated and if they will be retained (e.g. process memory pages).
```
paolonig, line 85: If the concern here is time interference I think the problem does not limit to resource retention only. I would say that, <<given the current complexity of HW SoCs and of the programs running on top, it is hard to claim that a safety workload is able to meet the allocated FDTI with a certain ASIL rigor without an external monitor>>

Contributions/Linux_Kernel_Safety_First_Principles.md
```diff
@@ -0,0 +1,141 @@
...
+
+
+## **First Principles - Availability**
+This section provides touchstone concepts about (lack of) system availability, in presence of non-safety-qualified components.
+
+1. Detecting interference alone doesn't help with controlling/managing availability.
+2. The nature of the very complex systematic interference makes estimating probability of a failure unrealistic.
```
paolonig, line 92: I disagree on this point. For instance even if an external safety monitor is required to meet the FDTI claim, it is still possible to tune the system and stress test it to claim that the unavailability risk due to temporal interference is acceptable

Contributions/Linux_Kernel_Safety_First_Principles.md
```diff
@@ -0,0 +1,141 @@
...
+3. Stress testing as a means of safety qualification might be realistic at most for very simple cases.
```
paolonig, line 93: But here the topic is availability, not safety qualification....

Contributions/Linux_Kernel_Safety_First_Principles.md
```diff
@@ -0,0 +1,141 @@
...
+4. Qualitatively, the likelihood of a failure grows with invocativation of non-safety-qualified components. For example:
+   1. invocation of device drivers, especially if not related to safety operations
+   2. syscalls
+   3. memory allocation/release operations
+   4. I/Os, e.g. storage, network
+   5. evaluation of Linux Security Module hooks
+   6. evaluation of cgroups tests
+   7. presence of non-safety-relevant triggering many of the above points
+
```
paolonig, line 102: Above points are correct, and this is why a QM baseline is always required to support availability, but the same baseline applies to NSR components....so what is the goal here? Wouldn't it be easier to say "a QMS baseline is required to meet the products' availability requirements" ?

Contributions/Linux_Kernel_Safety_First_Principles.md
```diff
@@ -0,0 +1,141 @@
...
+
+## **Considerations - Toward the safety goals**
+Considerations based on the two previous sections.
+
+1. Devise negative testing simulating interference toward safety-relevant components.
+2. Rely only on external components/mechanisms that already qualified at same-o-better safety level.
```
paolonig, line 108: rather than "external" I would say "independent"

Contributions/Linux_Kernel_Safety_First_Principles.md
```diff
@@ -0,0 +1,141 @@
...
+3. Analyse effects of interference based on recipient (e.g. corrupted data harder to notice)
```
paolonig, line 109: Here I would say "Make sure that the interference modes are avoided or detected either at the source or at the recipient side with sufficient level of rigor"

Contributions/Linux_Kernel_Safety_First_Principles.md
```diff
@@ -0,0 +1,141 @@
...
+4. When possible, leverage system configuration (e.g. external safety island / external monitor)
+5. Pick adequate granularity for verification (e.g. fine grained component validation vs end-to-end)
+6. In case of end-to-end verification, choose strategy: periodic test injection, or on demand.
```
paolonig, line 112: IMO point 6 is about "Make sure that mechanisms used to prevent or detect interference are properly verified. In case of HW mechanisms periodic or on-demand diagnostic tests may be required"

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: I think I'm stating the facts, while you are making the assumption that such evidence /argumentation exists
Burden of proof is on you, then. :-) 
Till the evidence is provided and verified, this is the situation.

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: A barrier is a barrier. A barrier is something that does not let the interference pass.
STACKPROTECTOR is an after-the-fact mechanism.
And notice that stackprotector doesn't protect anything, in practice.
"Stackdetectector" would be a better name.

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: unit is meant to represent the granularity chosen for the qualification, whatver that might be.
I'm trying to say SeooC without using the terminology, but that is the intended meaning.
As long as he meaning is preserved, I'm open to re-formulating it.

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: "Currently" is implied, but, again, as long as something is proven to work, it's an absolute statement.
It can be rephrased, if it helps with readability, without altering the meaning.
My point is that the "Very complex systematic corruption" cannot be detected through solutions that are applicable to - let's say - the memory map of a safety-relevant process, that might comprise tens, hundreds of MB of memory. You can certainly try to use ad-hoc verification, e.g. performing functional testing on that process, but it needs to be tailored to each individual process. No generic solution. Only a generic methodology, to be taylored.

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: No. Stress testing is certainly beneficial, but alone it is not sufficient.
Or else you are saying that QM practices can turn something into ASIL.
I'm sure you don't mean that.

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: It is trying exactly to draw a line between claims that are sound and claims that are not.
Of course one can make a leap of faith, cross the fingers and hope.
But without focused negative testing it is impossible to assess that safety claims on the kernel (alone) are valid.
Of course one can choose other mechanisms, e.g. redundancy, but that is a completely different system.
Once more, I'm happy to be proven wrong, but without verifiable counter evidence, the claim stands.

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: Isn't it obvious that there is an association?

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: The topic is the interference (AKA final bad effect).
The kernel doesn't really support anything, being QM at best.
Again, waiting for objective counter evidence.

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: simple -> direct ?

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: NSR = Non Safety Relevant? 
If that is the case, then no, I really mean qualified, as out of context.
cgroups are not qualified - and qualifying them would mean qualifying the entire kernel, more or less
same goes for LSM/SELinux


The fact that one might want to use them to prevent a certain type of interference dpoesn't automagically qualify them :-D

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: The problem is not with the HW, in fact here are systems which are qualified.
Put Linux on the same HW and this problem arises.
It's really an OS problem.

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: Feel free to estimate in a verifiable way said probability, then.
We can review the soundness of the estimation.

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: The point is that a QM component is by definition less safe than an ASIL one.
Even in presence of perfect detection of interference QM->ASIL, the more QM components can interfere with the ASIL one, the lower the MTBF.
A QM baseline might not be enough to meet availability requirements.

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: can you name an independent internal component?

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: No, it's intentionally excluding the source  because the whole point is that it's impossible to claim completeness about sources.

Again, if you disagree, feel free to prove me wrong and eenumerate all the possible sources.

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: this is not about HW - not necessarily

it's about using QM or less SW while performing E2E verification that it's still functioning correctly

Contributions/Linux_Kernel_Safety_First_Principles.md
reiterative:
```suggestion
4. Internal interference can also cascade through components that have only been safety-qualified at a component level.
```

Contributions/Linux_Kernel_Safety_First_Principles.md
reiterative: I think we can make a stronger (but narrower) statement here:
```suggestion
6. Complex systematic corruption of writable memory cannot be prevented by a generic, kernel-level solution
```

Contributions/Linux_Kernel_Safety_First_Principles.md
reiterative: I concur that both positive and negative testing are required. I would phrase it like this:
```suggestion
8. Where safety claims involving the Linux Kernel are supported by testing, both positive, requirements-based testing and focused, analysis-led negative testing are required.
```

Contributions/Linux_Kernel_Safety_First_Principles.md
reiterative:
```suggestion
10. If safety claims relating to one component are based on support by another, then the supporting component must have the same or better level of safety integrity (e.g. a safety-qualified watchdog).
```

Contributions/Linux_Kernel_Safety_First_Principles.md
reiterative: I think a stronger statement about the nature of the risk is clearer here:
```suggestion
11. Any other process could potentially interfere with a safety-qualified process, indirectly, through the kernel (e.g. a memory management bug could trigger interference between such processes).
```

Contributions/Linux_Kernel_Safety_First_Principles.md
reiterative:
```suggestion
15. When there are many possible sources of interference in a system, we can only reliably model and detect it in the receiving context.
```

Contributions/Linux_Kernel_Safety_First_Principles.md
reiterative:
```suggestion
16. The precise timing of allocation (e.g. during system or process startup) and retention of dynamically managed shared resources (e.g. process memory pages) is difficult to predict in a Linux-based system
```

Contributions/Linux_Kernel_Safety_First_Principles.md
paolonig: @reiterative  I would rephrase as <<In a mixed-criticality scenario, all unqualified code represents a potential risk of interference, which grows with frequency of execution; therefore any code supporting the mixed criticality safety claim (e.g. cgroups/containers, LSM/SELinux, etc.) needs to be qualified according to the highest safety integrity level of the supported safety processes>>

The reason being that the way it is currently written sounds like "cgroups/containers, LSM/SELinux" are bad actors; instead I think we should focus on what needs to be done if we want to leverage such features...

Contributions/Linux_Kernel_Safety_First_Principles.md
paolonig: @igor-stoppa  maybe a specific integrator decides that the associated performance drop is OK...you dunno 100% for sure, therefore @reiterative  I would rephrase as 

<<When there are many possible sources of interference in a system, it is usually easier to model and reliably detect it in the receiving context>>

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: @reiterative I'm almost ok with this, but I do not like that "only" because it seems to imply that something can be done on components to prevent the cascading of interference.
In theory yes, you could add redundance and what not, but in practice I don't think anyone will be able to make components sufficiently hardened.
If that is what you are after, then let's spell it out. I want to paint an accurate picture of the problem, without closing avenues that are open, but I do not want to present a rocky mountain path as if it was a boulevard either :-D

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: @reiterative: "neither prevented nor detected" is the right formualtion, I think

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: But they *ARE* bad actors, since they are not ASIL
Again, evidence supporting the opposite is welcomed.
Till then, I think I stated the existing situation.

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: and, again, it's not about performance, it's about completeness: you need to present not only the sources covered, but also the ones NOT covered, to claim a certain level of coverage.
And it's the latter part that is practically impossible.
Again, if it is possible, feel free to prove me wrong.

Contributions/Linux_Kernel_Safety_First_Principles.md
paolonig: Then any piece of the Kernel is a bad actor unless you find all the evidence that you need to build a safety case.
IMO the goal of this doc should be to help focusing on the challenges that must be solved for using the Kernel in supporting certain safety claim. So if you need to claim mixed criticality between user space processes you need to focus on the qualification of MM, Scheduler, namespaces, cgroups, SELinux etc. 
Instead if you need to claim FFI between Kernel threads (that also includes interference of a QM driver towards a user space process), you still need to focus on the qualification of MM and scheduler plus you need to model the interference from a QM driver toward an ASIL kernel component and toward a user mode thread.

Contributions/Linux_Kernel_Safety_First_Principles.md
paolonig: You can do that only if you define what is a good-enough RT performance. BTW you can start from [this paper](https://bristot.me/files/research/papers/ecrts2020/deOliveira2020Demystifying.pdf) and the toolset mentioned within it as a good starting point.

Contributions/Linux_Kernel_Safety_First_Principles.md
paolonig: Most of the use cases I worked on did not require the OS to be safe operational. I.e. panicing the OS would lead the system in an acceptable safe state (with a very upset driver, hence the need for availability, that however is not related to safety)

dweingaertner (commented):
One of the main issues being discussed is the ability the kernel has to change any part of memory. Does someone know how such problem is dealt with in other OS which have high SIL certifications? For example, when using a hypervisor? Is the argumentation based on the simplicity of the hypervisor`s code and expert knowledge evaluating it?

Contributions/Linux_Kernel_Safety_First_Principles.md
paolonig: Sorry but you are wrong here. A detected interference is not a failure (for fail safe systems) of the system and hence it does not contribute to the final MTBF target. For safe operational then the story is different as detecting the interference is not a good-enough mechanism (but then you need to disambiguate this clearly as usually availability is considered separate from safety  )

Contributions/Linux_Kernel_Safety_First_Principles.md
paolonig: The combination of a Runtime Verification Monitor and an external WTD can be used as an independent monitor for example.

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: Indeed: there is no reason to expect any safety level from any prt of the kernel, unless proven differently.

The goal of this document is simply to provide *facts* that one can start from.
Speculations can certainly be made available, but not here.

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: I'm not talking about RT performance, at least not in the time scale commonly intendend. 
I'm talking about minutes (or more) overheads to basic operations.

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: Ok, then just keep the system off :-)

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: anyways the RVM out of the box doesn't provide that functionality, you have to feed it the models

Contributions/Linux_Kernel_Safety_First_Principles.md
paolonig: @igor-stoppa  I think your statement is not correct. For example CONFIG_STACKPROTECTOR_STRONG or KASAN are generic Kernel solution that provide detections of writable memory. So I would stick with Pauls' rewording

Contributions/Linux_Kernel_Safety_First_Principles.md
paolonig: Can you make an example of a basic operation and also explaining what is preventing it from running?

Contributions/Linux_Kernel_Safety_First_Principles.md
paolonig: Jokes apart I think that in order to agree to such principles we need first to define the integration context.
As I mentioned all the customers I worked with usually accept to have an availability test campaign to be separate from safety (since safety is guaranteed by the external monitor in case of missing the FDTI deadline)

Contributions/Linux_Kernel_Safety_First_Principles.md
paolonig: As above it all depends on your system, on your integration context. 
If for the integrator it is safety critical to avoid interference, since the only detection would still lead to a hazard, then you're right.
If instead detecting interference triggers the safe state that relies on a backup system, for example, pulling over the car, then you're not right.
So I think we should better clarify the assumptions on the integration context...

Contributions/Linux_Kernel_Safety_First_Principles.md
paolonig: Yes of course, you need to design your specific RVM. It was just an example to justify independent as a better word compared to external....

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: @paolonig They are partial solutions. Placing canaries on the stack doesn't catch all the possible corruptions.
KASAN alters the system configuration, and anyways is not deployed.
Feel free to provide evidence countering my arguments, they are intentionally made in a very clan cut way, to allow easy confutation by evidence, if any can be found.

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: Imagine a set of processes implementing machine vision using tens of GB or ram.
Try to monitor that in realtime for corruption caused by the kernel ...

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: These principles are meant to not be democratically decided by what anyone customer's might think.
I understand "the customer is always right" approach, but that doesn't belong to this document.

Anyone is free to take the risks they prefer, small or large.

This document is not about that. It's about separating assessments that can be done with very high confidence, if not certainty, from others that are based on subjective analysis.

For example, can any of your customers prove that they have determined the coverage of that stress testing?

Performing the same or similar operation billions of times doesn't say anything about other scenario that might arise.
It can be done for sure in simple scenarios, but in more complex ones it cannot.
Again, I'm tryin to make it easy to prove me wrong, by making fairly clean cut statements.

Just add evidence ...

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: The backup system is out of the scope of this document.
You can very well consider it as confirmation of my point.
This is talking about availability of the intended functionality of the "product", but let's agree it to be a car.
A pull over is a failure. It's not a safety failure, certainly. But it IS a failure of the product.
We can add a clarification about what MTBF means in this context (what I just wrote), if it can help.

Contributions/Linux_Kernel_Safety_First_Principles.md
paolonig: But then rather than e2e verification it seems an actual ASIL decomposition (and in this case we need to make sure that an appropriate safety analysis is done on the QM element to make sure the ASIL monitors cover all the dangerous failure modes)....

Contributions/Linux_Kernel_Safety_First_Principles.md
paolonig: Yes they are partial generic kernel level solutions, I agree. However an integrator or a SW vendor may use such partial solution in combination with other measures to support a non-interference claim.
So IMO the fact is the monolithic nature of the Kernel that does not allow, today, interference prevention (like the one you can have by SMMU for a user space process) and we can all agree on this.
What I cannot agree is about principles that also imply or deny specific ways to support a certain claim; since the specific ways depend on the specification of the integration context and of other projects aspects (depending on the engineering workforce I may decide to use some verification strategies rather than others....)

Contributions/Linux_Kernel_Safety_First_Principles.md
igor-stoppa: if one end can make asusmptions about the other end, and validate them for correctness, isn't it e2e?

igor-stoppa:
I can't reply directly to @dweingaertner 's question, but here it is:

> One of the main issues being discussed is the ability the kernel has to change any part of memory. Does someone know how such problem is dealt with in other OS which have high SIL certifications? For example, when using a hypervisor? Is the argumentation based on the simplicity of the hypervisor`s code and expert knowledge evaluating it?

They are usually designed from ground-up to not be exposed so much to this problem.
Linux is monolithic, others are microkernels. A typical device driver is not allowed write access to kernel vital data, in a microkernel. But in Linux it can, and that it one problem. Having in-kernel barriers, then it becomes easier to qualify one the relatively small portion of code that has high privilege.
And usually the develoment follows different processes, and there is more control on what gets accepted and why and how.

xen and zephyr are perhaps the best examples of FOSS that is developed in a safety-friendly way.

Regarding the hypervisor:
The hypervisor knows about  an OS as much as the OS knows about one of the preocesses it runs.
Very little. And intentionally so.

Furthermore, notice that for an hypervisor to realistically poke its nose in the kernel and help with integrity, either the kernel acts cooepratively, abstracting its own data structures through specificly crafted hypervisor calls, or the hypervisor needs to have knowledge of the kernel internals, using its headers and probably libraries.
Many wnat to have a closed source hypervisor and that would not be possible, given the contagious nature of the GPLv2 used by the kernel.

These are just a few reasons, but they should prove to be already sufficient :-)

PR #37, opened 2024-07-12
Update Using_Linux_in_a_Safe_System.md

igor-stoppa:
Added System Decomposition section

### DIFF
--- a/Contributions/Using_Linux_in_a_Safe_System.md
+++ b/Contributions/Using_Linux_in_a_Safe_System.md
@@ -412,6 +412,64 @@ and/or costly.
     of the SoC watchdog toward multiple operating systems, however it
     could go beyond that, and reach higher levels of complexity.
 
+  * **System Decomposition**
+ 
+    The previous points can be leveraged and extended to cases where the
+    safety requirements rely on multiple subsystems, possibly across different
+    software layers.
+    
+    As usual, it is extremely important to assess upfront the requirements
+    for availability.
+
+    A requirement for simply detecting the potential interference can be easier
+    to deal with, than a requirement to ensure continued safety operations,
+    in the presence of interference.
+
+    Leaving aside, for now, considerations about the latter scenario, the former
+    one still carries a need for the detection to happen within a maximum interval,
+    measured from the beginning of the interference.
+
+    With these two requirements in mind:
+    * detecting interference
+    * detecting the interference within a maximum allowed time
+      
+    One can attempt to isolate those subsystems directly involved, and investigate
+    the possibility of periodically validating its continued integrity.
+    
+    Here, the basic concept is that one could be able to provide - periodically -
+    a preset stimuli, and compare the live output of the system with the
+    pre-calculated expected one, that the system should produce, when free from
+    interference.
+    
+    Taking as example an image recognition pipeline, tasked with receiving an input
+    from a sensor and recognising items from a given set (e.g. a speed limit sign),
+    this could be implemented in the following way:
+    * periodically inject a known image, of a specific object(s)
+    * expect that within a given time the system will correctly identify the object(s)
+    
+    Such an approach has the advantage of not requiring any special partitioning of
+    the architecture (e.g. which parts must be implemented as user processes and
+    which parts must be kernel components).
+    
+    However, to work, it must also satisfy several conditions:
+    * the monitoring must happen form a context that is at least of equal ASIL level
+    * the monitoring context must not be affected by the interference it is supposed to detect
+    * the testing must have sufficient coverage for detecting all the relevant types of inteference
+    * the testing must account for any recently collected hystorical data
+      that the system might rely on.
+      (e.g. lighting statistics from past frames might get corrupted and affect
+      the following evaluations, but if they are not part of the periodic self test,
+      this corruption wil lgo unnoticed, because the system tested is different from
+      the system in real use)
+
+    The requirements of the Fault Deteting Time Interval (FTDI) and the time required
+    for performing a self-test will dictate how often this self test must be performed.
+
+    Should it be required to ensure availability, it could be possible to use multiple
+    instances of the detection system, as long as it can be proven that they are
+    sufficiently decoupled to not be victim of the same interference, simultaneously.
+    
+    
 ## **Hardware features are not a replacement for good design**
 
 ***Against stupidity the gods themselves contend in vain.***


Contributions/Using_Linux_in_a_Safe_System.md
```diff
@@ -412,6 +412,64 @@ and/or costly.
...
+    from a sensor and recognising items from a given set (e.g. a speed limit sign),
+    this could be implemented in the following way:
+    * periodically inject a known image, of a specific object(s)
+    * expect that within a given time the system will correctly identify the object(s)
+    
+    Such an approach has the advantage of not requiring any special partitioning of
+    the architecture (e.g. which parts must be implemented as user processes and
+    which parts must be kernel components).
+    
+    However, to work, it must also satisfy several conditions:
```
paolonig, line 454: IMO there is a critical aspect that we are missing here:
For ASIL decomposition to be valid we need to have:
a) Redundant requirements allocated to both the monitored subsystems and the monitors
b) A safety analysis on the monitored subsystems shall verify the completeness and sufficiency of the safety requirements allocated to the monitor and shall also verify the adequacy of the monitor design
c) The monitored subsystems and the monitors must be independent

In the text below you are basically elaborating on point c) while you're missing to elaborate on point a) and b). The main issues being that the higher the complexity or the decomposed requirements and of the monitored code, the higher the complexity and cost of the monitor is. This is an important consideration to be made for choosing or not choosing to afford the redundant monitor design, verification and deployment costs...

Contributions/Using_Linux_in_a_Safe_System.md
```diff
@@ -412,6 +412,64 @@ and/or costly.
...
+    * the testing must have sufficient coverage for detecting all the relevant types of inteference
+    * the testing must account for any recently collected hystorical data
+      that the system might rely on.
+      (e.g. lighting statistics from past frames might get corrupted and affect
+      the following evaluations, but if they are not part of the periodic self test,
+      this corruption wil lgo unnoticed, because the system tested is different from
+      the system in real use)
+
+    The requirements of the Fault Deteting Time Interval (FTDI) and the time required
+    for performing a self-test will dictate how often this self test must be performed.
```
paolonig, line 466: I would not use the word "test" here. In general the last part of an ASIL Decomposition flow is a "comparator" comparing the outputs from the monitor and the monitored system respectively.
Also you should not confuse the FDTI with the comparator sampling interval. You may have also an analog continuous comparator that trigger an exception as soon as the delta reaches a certain threashold. In this regard the FDTI is the max time to handle the exception...

Contributions/Using_Linux_in_a_Safe_System.md
```diff
@@ -412,6 +412,64 @@ and/or costly.
...
+
+    Should it be required to ensure availability, it could be possible to use multiple
+    instances of the detection system, as long as it can be proven that they are
+    sufficiently decoupled to not be victim of the same interference, simultaneously.
```
paolonig, line 470: This is not correct. If Linux, in this case the monitored entity, is the only "channel" driving the actuator, you can have as many monitors as you like but you'll never be able to recover from a wrong actuator output.
For this to happen you need a 2oo3 architecture where at least 2 channels are capable of producing redundant actuator output and a monitor is able to tell which of the two channel is broken in case of compare mismatch and hence able to pick the output from the sane channel

Contributions/Using_Linux_in_a_Safe_System.md
```diff
@@ -412,6 +412,64 @@ and/or costly.
...
+  * **System Decomposition**
+ 
+    The previous points can be leveraged and extended to cases where the
+    safety requirements rely on multiple subsystems, possibly across different
+    software layers.
+    
+    As usual, it is extremely important to assess upfront the requirements
+    for availability.
+
+    A requirement for simply detecting the potential interference can be easier
```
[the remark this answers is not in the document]
reiterative: @paolonig ASIL Decomposition is an example of applying the principles of system decomposition with a particular set of goals. I believe this document is intended to describe the general approach for system decomposition, rather than ASIL decomposition in particular.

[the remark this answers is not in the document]
paolonig: @reiterative  fine, I am not suggesting to use the "ASIL Decomposition" wording here, I am just highlighting that we are only focusing on the monitoring of interference failure modes while the monitoring can be used also to detect failure modes deriving from a low systematic capability of the monitored subsystems...

Contributions/Using_Linux_in_a_Safe_System.md
igor-stoppa: Costs, intended as monetary costs, are subjective, and likely to change over time, I'd rather leave them out entirely.
Besides, costs can be optimised later on, when the goal is somehow achieved.

Contributions/Using_Linux_in_a_Safe_System.md
igor-stoppa: I meant to indicate an upper bound on the priodicity. I will reword the text to make it more explicit. Then the analog comparator will be automatically included (although even that is not really istantaneous, it will have a specific delay of propagation - nothing is instantaneous).

Contributions/Using_Linux_in_a_Safe_System.md
igor-stoppa: ok for the numbers, I wrote multiple, but that would have included also 2, which is not sufficient

regarding the actuator, I had in mind exactly what you described, meaning that the acutator receives the vetted output, however I agree that it was not as clear as it coudl have been - I will reword that too

Contributions/Using_Linux_in_a_Safe_System.md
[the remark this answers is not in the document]
igor-stoppa: I interpret this comment as not requiring changes.

Contributions/Using_Linux_in_a_Safe_System.md
igor-stoppa: Perhaps we should have a separate section or even document about costs, if it is an aspect that is considered within the scope of ELISA. I've seen it surfacing in many cases. I'm not oblivious to it, but markets can be so very different that talking about costs can became a very tricky subject. @reiterative @paolonig I gladly leave that to the both of you ;-)

Contributions/Using_Linux_in_a_Safe_System.md
igor-stoppa: done

Contributions/Using_Linux_in_a_Safe_System.md
igor-stoppa: done

reiterative (approved):

reiterative:
@paolonig Please can you review and confirm that your points have been addressed?

PR #38, opened 2024-07-12
Create Linux_Memory_Management_Essentials.md

igor-stoppa:
Description of internal memory management in Linux, from safety perspective.

### DIFF
--- a/Contributions/Linux_Memory_Management_Essentials.md
+++ b/Contributions/Linux_Memory_Management_Essentials.md
@@ -1,0 +1,242 @@
+# **Linux Memory Management Essentials (Work In Progress)**
+
+## Index
+
+[Terms and Abbreviations](#Terms-and-Abbreviations)
+
+[References](#References)
+
+[Disclaimer](#Disclaimer)
+
+[Purpose of the document](#Purpose-of-the-document)
+
+[Structure of the document](#Structure-of-the-document)
+
+[Kernel-space memory allocations](Kernel-space-memory-allocations)
+
+[User-space memory allocations](User-space-memory-allocations)
+
+[License: CC BY-SA 4.0](#License-CC-BY-SA-40)
+
+
+## **Terms and Abbreviations**
+Plese refer to the Linux Kernel documentation.
+   
+
+## **References**
+
+1. [Linux Kernel website](https://www.kernel.org) - <https://www.kernel.org>
+2. ***Interference Scenarios for an ARM64 Linux System***
+3. [CC BY-SA 4.0 Deed | Attribution-ShareAlike 4.0 International | Creative Commons](https://creativecommons.org/licenses/by-sa/4.0/) - <https://creativecommons.org/licenses/by-sa/4.0/> License
+
+
+## **Disclaimer**
+* This document is not intended to be a replacement for understanding the memory management of the Linux Kernel,
+  nor it attempts to be an exhaustive analysis of safety implications.
+* Because of the very volatile nature of the code within the Linux Kernel, each of the statements made
+  below should not be taken at face value, but rather verified, for any Linux kernel version following
+  the one used while writing the document (6.9).
+* When referring to specific HW features, the document refers to the ARM64 architecture.
+
+## **Purpose of the document**
+This document aims to provide an holistic view of what happens in Linux memory management, so that
+one is at least aware of certain features and can use this document as a jumping pad toward more detailed documentation.
+Or even to the code base.
+
+## **Structure of the document**
+The document is divided in two parts, based on the destinatary of the memory allocations discussed: kernel-space or user-space.
+Individual points are numbered for ease of reference, but the numbering is not meant to represent any sequence.
+
+## **Memory management in Linux**
+
+### **Kernel-space memory allocations**
+
+#### **Directly Verifiable Assertions**
+The following section presents a set of statements that can be objectively verified e.g. by inspecting the sources.
+
+1. unlike processes memory, kernel memory pages are not swapped, nor dropped silently by the kernel itself,
+   although an hypervisor will do to a VM what the kernel does to a process (but this is beyond the control of the kernel)
+2. the kernel context (usually EL1 on ARM64) uses one single memory map (page tables) across all the cores
+   executing in kernel mode
+3. on 64 bit systems (e.g. ARM64 and x86_64), usually almost all physical memory is mapped in a
+   (semi)contiguous (there can be holes) range. Memory within this range is both virtually and physically contiguous.
+4. physically contiguous memory is treated as a scarce resource, and typically is not provided to userspace,
+   unless it explicitly asks for it (e.g. for implementing some DMA buffer)
+5. the kernel can access userspace memory in three ways:
+   1. through the userspace mapping 
+      1. this type of access is limited to few functions, like copy_to_user()/copy_from_user() and put_user()/get_user()
+      2. outside the execution of the functions mentioned, this type of access is not possible, because the userspace mappings are made available only while executing such functions
+      3. the userspace memory map can implement HW protections against being misused by the kernel
+      4. the kernel is able to access process pages in the same sequence and with the same mappings as the process does
+   2. through a memory buffer (e.g. a memory area where the kernel regularly needs to perform large amount of read/write operations, like a network buffer)
+      1. the memory used by a buffer is well defined and delimited, rather than a generic area, and it is specifically reserved for this purpose
+      2. the kernel uses own mappings in EL1, while the user space uses the EL0 ones
+      3. very often the kernel doesn't actually access this area directly, but it rather configures a DMA controller to do the transfers, directly to the physical memory.
+      4. misconfiguration of these peripherals that can access physical memory directly is a potential problem - a form of mitigation relies on using IOMMUs: hardware components that the kernel can configure as a firewall on the phisical memory bus, to limit what memory is accessible to each memory bus master device.
+   3. through the EL1 mappings + linear map
+      1. not the intended way for the kernel to access process context, because the intended way is one of the two previously described
+      2. attempting to use EL1 mappings would be not very useful, both for legitimate purposes and even for malicious ones:
+         1. the sequence of pages mapped in the userspace process is not known, and it can change continuously, as the memory management does its job of optimising page allocation to running processes
+         2. some process pages might even be "missing", because they have been either swapped out or dropped (see previous point)
+         3. even for an attacker, in a security scenario, because most likely the attacker would need to access user space pages sequentially, or anyway through user mappings (this is why EL1 mappings are allowed to have access to pages containing EL0 code/data: the security risk is relatively low)
+         4. it bypasses any protection that the process might employ through its own mapping, which, even for legitimate kernel operations on userspace, would be less safe
+6. Assertions about page "mobility", from different perspectives:
+   1. employment of physical pages - observing a physical memory page (e.g. a 4kB chunk aligned on a 4kB boundary) and how it is employed: what sort of content it might host, over time.
+      1. certain physical pages are put to use for a specific purpose, which is (almost) immutable, for the entire duration of the execution of the kernel:
+         1. kernel code from the monolithic portion (except for pages containing kernel init code, which are released after kernel init is completed)
+         2. kernel statically allocated data from the monolithic portion (except for pages containing kernel init data, which are released after kernel init is completed)
+         3. some kernel dynamically allocated data, used by the kernel itself, and never released, due to the nature of its use (e.g. object caches, persistent buffers, etc.)
+         4. memory used for loadable kernel modules (code, data) is tied to the permanence of the module in memory - this is typically stable through the entire duration of the execution. Some exceptions are modules loaded/unloaded as a consequence of certain peripherals (dis)appearing, e.g. USB ones.
+      2. other physical pages (most, actually, in a typical system) are available for runtime allocation/release cycles of multiple users:
+         1. transient kernel linear memory allocations (kmalloc / get_free_pages)
+         2. transient kernel virtually linear memory allocations (vmalloc for kernel address space)
+         3. user-space memory allocations (they are always virtually linear), which are by default transient and exposed to repurposing proactively done by the kernel (see below)
+   2. transitioning of a logical page - given certain context and content, where it might be located in memory over time - and if it might be even discarded.
+      1. the kernel doesn't spontaneously repurpose the utilisation of its own physical pages; therefore it is possible to assume that the logical content of kernel allocations will reamin tied to the associated physical pages, as long as it is not intentionally altered (or subject to interference)
+         1. metadata memory used by the kernel for housekeeping purposes related to processes is included in this category; examples: task/cred structures, vmas, maple leaves, process page tables.
+      2. memory used by the kernel for the actual userspace processes: the content of this logical page determines its life cycle and expectancy: certain content such as code or constants can be "re-constructed" by relaoding it from file (code pages are likely to necessitate undergoing re-linking), so the actual logical content might disappear, over time. Other pages, on the other hand, are meant to hold non-re-constructible content, such as stack, heap, and variable data. These pages can, at most, be swapped out, and loaded back later on, but they cannot be simply dropped.
+      3. page cache: it is a collection of memory pages containing data read from files, over time; e.g. code or initialised data from a file that was accessed recently; in some cases the page might never have been used, but it was loaded as part of the readahaead optimisation. The life expectancy of these logical pages is heavily affected by how many processes might keep accesing them and the level of memory starvation of the system caused by other processes, with some additional complexity layered on top of this, by the use of containers.
+   3. The kernel utilises various optimisations that are meant to take advantage of hardware features, such as multi-stage caching, and also to cope with different memory architectures (like Non Unifor Memory Architecture - NUMA). The main goals are:
+      1. avoid having to propagate too frequently write operations through the layers of HW cache, which is caused by pages being evicted from the cache, due to memory pressure
+      2. avoid having multiple cpus writing to the same page, in a NUMA system, where only one cpu has direct write access to that memory page, because it would cause cache invalidation
+      Therefore, the kernel tends to:
+      1. reuse as much as possible a certain page that has just been freed (so called hot page, since it is stll presumably present in the HW cache)
+      2. keep for each core a stash of memory pages readily available (which prevents other cores from accessing said pages and introducing additional cache-flush operations)
+   4. The MMU, involved in performing address translations, acts also as a bus master, and performs read operations whenever it needs to do an address translation that is not already present in its own local translation cache (TLB - Translation Lookaside Buffer). Having to perform too many of such address translations (page walks) can constitute a significant performance penalty. The TLB is not very large, and accessing lots of different memory addresses that do not belong to the same translation entry can cause severe performance degradation. This is why the kernel actually keeps most of the memory mapped in the linear map, to take advantage of a feature present in many processors, that allows the mapping of large chunks of physical memory (e.g 2MB) as a single entry (or few ones). The kernel code, for example, is kept compact to maximise the efficiency of the fetching operations. For the similar reasons, the linear mapping allows to have memory already mapped, without a need for creating those mappings on the fly.
+   5. For what concerns allocation from the linear map (kmalloc / get_free_pages), the kernel attempts to keep the free pages as much continuous as possible, avoiding fragmentation of the free areas.
+      1. this is implemented through the concept of the "buddy allocator", meaning that whenever a certain amount of linear memory is requested (either sub-page or multi-page size), it always tries to obtain it from the smallest free slot available, only breaking larger free slots when no alternatives are possible.
+      2. the kernel also keeps ready a certain amount of pre-diced memory allocations, to avoid incurring the penalty of having to look for some free memory as a consequence of an allocation request.
+      3. folios are structures introduced to simplify the management of what has been traditionally called compound pages: a compound page represents a group of contiguous pages that is treated as a single logical unit. Folios could eventually support the use of optimisations provided by certain pages (e.g. ARM64 allows the use of a single page table entry to represent 16 pages, as long as they are physically contiguous and aligned to a 16 pages boundary, through the "contiguous bit" flag in the page table). This can be useful e.g. when keeping in the page cache a chunk of data from a file, should the memory be released, it could result in releasing several physically contiguous pages, instead of scattered ones.
+   6. whenever possible, allocations happen through caches, which means that said caches must be re-filled, whenever they hit a low watermark, and this re-filling can happen in two ways:
+      1. through recycling memory as it gets freed: for example in case a core is running short of pages in its own local queue, it might "capture" a page that it is freeing.
+      2. through a dedicated thread that can asynchronously dice larger order pages into smaller portions that are then placed into caches in need to be refilled
+   7. the kernel can also employ an Out Of Memory Killer feature, that is invoked in extreme cases, when all the existing stashes of memory have been depleted: in this case the killer will pick a user space process and just evict it, releasing all the resources it had allocated. It's far from desirable, but it's a method sometimes employed.
+   8. freeing of memory pages also happens in a deferred way, through separate threads, so that there is no overhead on the freer, in updating the metadata associated with the memory that has just been released.
+   9. All of the mechanisms described above for memory management are not only memory providers, but memory users as well, because they rely on metadata that cannot be pre-allocated and must be adjusted accordingly to the memory transactions happening, as the system evolves over time; therefore they also consume memory, generating an overhead, and being exposed themselves to interference.
+   10. The Linux kernel provides means to limit certain requests a process might present; for example with cgroups it is possible to create logical memory "bubbles" that cannot grow beyond a set size, and associate processes to them, that share the collective limit within the bubble. But this does not much toward separating how the kernel might use the underlying memory, besides setting the constraint as described.
+      
+#### **Safety-Oriented consideration**
+The following considerations are of a more deductive nature.
+
+1. Because of the way pages, fractions and multiples of them are allocated, freed, cached, recovered, there is a complex interaction between system components at various layers.
+2. Even using cgroups, it is not possible to segregate interaction at the low level between components with differnet level of safety qualification (e.g. a QM container can and most likely will affect the recirculation of pages related to an ASIL one)
+3. Because of the nature of memory management, it must be expected that memory management mechanisms will  interfere with safe processes, either due to a bug or due to the interference toward the metadata they rely on. For example, the memory management might hand over to a requesting entity a memory page that is currently already in use either by a device driver or by a userspace process playing a role in a safety use case.
+4. Still due to te complex interaction between processes, kernel drivers and other kernel code, it is practically impossible to qualify the kernel as safe through positive testing alone, because it is impossible to validate all the possible combinations, and it is equally impossible to assess the overall test coverage and the risk associated with not reaching 100%. The only reliable way to test is to use negative testing (simulating a certain type of interference) and confirming that the system response is consistent with expectations (e.g detect the interference, in case of ASILB requirements). And even then, the only credible claim that can be made is that, given the simulated type of interference, on the typology of target employed, the reaction is aligned with the requirements. Other types of targets will require further ad-hoc negative testing.
+5. Linux Kernel mechanisms like SELinux and cgroups/containers do not offer any protection against interference originating from the kernel itself.
+6. The Linux Kernel must be assumed to be QM, unless specific subsystems are qualified through both positive and negative testing.
+7. Even after achieving the ability to make certain claims about the Kernel integrity (or detection of its loss), this doesn't guarrantees system availability: knowing that an interference has occurred doesn't help with ensuring a certain level of system availability.
+
+### **User-space memory allocations**
+
+#### **Assertions**
+The following section presents a set of statements that can be objectively verified e.g. by inspecting the sources.
+
+1. given a user-space logical memory page, at a certain moment, it must not be assumed to have a corresponding
+   backing physical memory pages; it means that a process might have had some of its pages dropped and/or
+   swapped out to disk, or perhaps not even ever loaded.
+2. it is not possible to make many assumptions about the state of the logical content of a process pages.
+   Perhaps one of the few statments that can be made is that the logical page containing code being executed
+   in a certain moment has also a physical backing, while it is being executed.
+   It might be possible to introduce some additional certainties, for example by pinning down some memory allocations
+   by marking the process as non-swappable. The presence of caching and other
+   optimisations like the zero-page make it very difficult to be assertive beyond this point.
+3. the management of memory pages associated with a process is handled through the process memory
+   map, which consists of several virtual memory areas, representing address ranges within the
+   process address space, that are put to some use.
+   They come in two flavors:
+   1. anonymous mappings:
+      1. process heap, stack
+      2. zero-initialised variables
+      3. vDSO / vvar (code and data provided by the kernel, to optimise the execution of frequently performed operations, like user space reading the time of day)
+   2. file-backed mappings:
+      1. constants
+      2. the main executable associated with the process
+      3. the dynamic linker (optional but typically used for ELF files)
+      4. linked libraries (e.g. glibc)
+4. when the kernel starts a process, it sets up mappings for all the virtual memory areas required to get it started,
+   but it doesnt actually allocate any memory: once the process is scheduled for execution, it will eventually be run 
+   for the first time, and as soon as the various areas are accessed, they will trigger page fault exceptions.
+      1. when an address is accessed for the first time, it might require that a memory page is allocated to host whatever
+         the associated content might be, but it is possible, especially when the process has just been started, that
+         even the associated page in he page table is missing.
+      2. based on both availability of free pages and type of content asosciated, the access might cause the process
+         to sleep; exaples:
+         1. no free pages are available (unusual but possible) and the kernel will have to try to obtain one, in one of
+            the ways described above
+         2. the content needs to be loaded from a file, and the operation is blocking, because the retrieval process is much slower.
+      Either way, it is difficult to know if/when a process is not going to generate any more faults, and it is very much not deterministic.
+6. as also decribed in the previous point, the kernel uses various optimisations for dealing with processes on-demand mapping:
+   1. a physical page is allocated and mapped to a process only when the process accesses it, otherwise it might not be present:
+      1. file backed pages are allocated/mapped only when read/written to
+      2. anonymous pages are allocated only when written to
+   2. zero page: when a page is known to be empty, it is not reserved and mapped; instead,
+      the kernel has one specific (dummy) memory page that is mapped tothe process as read-only
+   3. shared libraries:
+      when the same library pages are mapped by multiple processes, the library physical pages
+      are mapped as read-only into each process address space
+   4. copy-on-write:
+      1. when a library has own data, this is initially mapped as read-only and shared;
+         only when written to, then a separate physical page is reserved for each writing process (threads sahre the same)
+      2. same happens for data pages that were treated as zero-page, but then are written to.
+   5. folios (see also kernel section): data structures that try to better abstract compound pages and *might* also be used
+      to represent optimised contiguous pages on ARM64 (instead of mapping 16 entries, it is possible
+      to map a 16-pages chunk of physically contiguous memory, that is also aligned to a 16 pages
+      boundary, reducing TLB use).
+      1. A folio acts as intermediary between vma and lower level memory management
+      2. it might pre-allocate/map more pages than explicitly requested
+7. read-ahead: when asked to fetch data from disk, the kernel might attempt to optimise the operation,
+   reading more pages than requested, under the assumption that more requests might be coming soon
+ 
+
+#### **Safety-Oriented consideration**
+The following considerations are of a more deductive nature.
+
+1. a process that is supposed to support safety requirements should  not have pages swapped out / dropped / missing,
+   because this would introduce:
+   1. uncertainty in the timing required to recover the content, if not immediately available
+   2. additional risk, involving the userspace paging mechanisms in the fulfilling of the safety requirements
+   3. additional dependency on runtime linking, in case the process requires it, and code pages have been
+      discarded - reloading them from disk will not be sufficient
+2. The optimisations made by the kernel in providing physical backing to process memory make it very
+   questionable if it can be assessed when a (part of) a process memory content is actually present in the
+   system physical memory.
+3. by default, it is to be expected that a process will be exposed to various types of interference from the kernel:
+   1. some of a more bening nature, like dropping of pages, or not allocation of not-yet-used one
+   2. some limited in extent, but hard or even practicaly impossible to detect, like a rogue write to process physical memory
+   3. some of systemic nature, like some form of use-after free, where a process page is accidentally in use also by another component
+   4. some of indirect nature, like for example when the page table of the process address space is somehow corrupted
+4. again, because of the extremely complex nature of the system, positive testing is not sufficient, but it needs to
+   be paired also with negative testing, proving that it is possible to cope with interference and detect it, somehow.
+5. the same considerations made about integrity vs. avaialbility for the kernel are valid here too: detecting
+   interference doesn't help with keeping it under a certain threshold, and due to the complexity of the system,
+   it is not possible to estimate the risk reliably.
+   
+
+
+## **License: CC BY-SA 4.0**
+
+### **DEED**
+### **Attribution-ShareAlike 4.0 International**
+
+Full License text: <https://creativecommons.org/licenses/by-sa/4.0/>
+
+**You are free to:**
+
+* **Share** — copy and redistribute the material in any medium or format for any purpose, even commercially.
+
+* **Adapt** — remix, transform, and build upon the material for any purpose, even commercially.
+
+The licensor cannot revoke these freedoms as long as you follow the license terms.
+
+**Under the following terms:**
+
+* **Attribution** — You must give appropriate credit , provide a link to the license, and indicate if changes were made . You may do so in any reasonable manner, but not in any way that suggests the licensor endorses you or your use.
+
+* **ShareAlike** — If you remix, transform, or build upon the material, you must distribute your contributions under the same license as the original.
+
+* **No additional restrictions** — You may not apply legal terms or technological measures that legally restrict others from doing anything the license permits.
+
+**Notices:**
+
+You do not have to comply with the license for elements of the material in the public domain or where your use is permitted by an applicable exception or limitation .
+No warranties are given. The license may not give you all of the permissions necessary for your intended use. For example, other rights such as publicity, privacy, or moral rights may limit how you use the material.
--- a/Contributions/README.md
+++ b/Contributions/README.md
@@ -7,6 +7,9 @@ ELISA umbrella.
 ## Index
 These markdown documents are actively extended and reviewed, as part of safety-related work done within ELISA.
 
+* [Linux Memory Management Essentials](Linux_Memory_Management_Essentials.md)
+  (3rd party) Summary of Memory Management features of the Linux Kernel that are also relevant for safety.
+  
 * [Using Linux in a Safe System](Using_Linux_in_a_Safe_System.md)
   (3rd party) Non-exhaustive list of engineering considerations and practices, that can help with designing a Safe System containing Linux.
   It can be seen as a companion to the Checklist below.


Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+Individual points are numbered for ease of reference, but the numbering is not meant to represent any sequence.
+
+## **Memory management in Linux**
+
+### **Kernel-space memory allocations**
+
+#### **Directly Verifiable Assertions**
+The following section presents a set of statements that can be objectively verified e.g. by inspecting the sources.
+
+1. unlike processes memory, kernel memory pages are not swapped, nor dropped silently by the kernel itself,
```
paolonig, line 57: should we add that kernel memory swapping can happen during PM Suspend ?

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+2. the kernel context (usually EL1 on ARM64) uses one single memory map (page tables) across all the cores
+   executing in kernel mode
+3. on 64 bit systems (e.g. ARM64 and x86_64), usually almost all physical memory is mapped in a
+   (semi)contiguous (there can be holes) range. Memory within this range is both virtually and physically contiguous.
+4. physically contiguous memory is treated as a scarce resource, and typically is not provided to userspace,
+   unless it explicitly asks for it (e.g. for implementing some DMA buffer)
+5. the kernel can access userspace memory in three ways:
+   1. through the userspace mapping 
+      1. this type of access is limited to few functions, like copy_to_user()/copy_from_user() and put_user()/get_user()
+      2. outside the execution of the functions mentioned, this type of access is not possible, because the userspace mappings are made available only while executing such functions
```
paolonig, line 68: I would say "outside the execution of the functions mentioned, this type of access would be an unexpected interference, because...", because to make it "not possible" you need HW protections as the ones you mentioned below

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+      3. the userspace memory map can implement HW protections against being misused by the kernel
```
paolonig, line 69: maybe better to mention ARM’s PXN/PAN explicitly here....?

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+      4. the kernel is able to access process pages in the same sequence and with the same mappings as the process does
```
paolonig, line 70: the point above is not clear to me....what do you mean? the sequence of accessing memory is driven by the code itself. A user space code cannot be kernel space code, also user space mapping (virtual mappings) are different than Kernel space ones....it is not clear to me what you intend to say...

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: I'd rather say here "dynamically swapped" and then have a separate comment about suspend, which is a valid topic but happens under different conditions.

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+   2. through a memory buffer (e.g. a memory area where the kernel regularly needs to perform large amount of read/write operations, like a network buffer)
+      1. the memory used by a buffer is well defined and delimited, rather than a generic area, and it is specifically reserved for this purpose
+      2. the kernel uses own mappings in EL1, while the user space uses the EL0 ones
+      3. very often the kernel doesn't actually access this area directly, but it rather configures a DMA controller to do the transfers, directly to the physical memory.
```
paolonig, line 74: Why are you thinking of DMA buffers? You could also have non-dma-able  phisical memory mapped both to user virtual address space as well as to Kernel virtual address space. I would consider the hazard associated with dma buffers as a separate additional point

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+      4. misconfiguration of these peripherals that can access physical memory directly is a potential problem - a form of mitigation relies on using IOMMUs: hardware components that the kernel can configure as a firewall on the phisical memory bus, to limit what memory is accessible to each memory bus master device.
+   3. through the EL1 mappings + linear map
```
paolonig, line 76: I would remove this from this list (since it is not a design mechanisms to be used to access user mode memory) and add a note (note that can be used later for the safety considerations)

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+         2. some process pages might even be "missing", because they have been either swapped out or dropped (see previous point)
+         3. even for an attacker, in a security scenario, because most likely the attacker would need to access user space pages sequentially, or anyway through user mappings (this is why EL1 mappings are allowed to have access to pages containing EL0 code/data: the security risk is relatively low)
+         4. it bypasses any protection that the process might employ through its own mapping, which, even for legitimate kernel operations on userspace, would be less safe
+6. Assertions about page "mobility", from different perspectives:
+   1. employment of physical pages - observing a physical memory page (e.g. a 4kB chunk aligned on a 4kB boundary) and how it is employed: what sort of content it might host, over time.
+      1. certain physical pages are put to use for a specific purpose, which is (almost) immutable, for the entire duration of the execution of the kernel:
+         1. kernel code from the monolithic portion (except for pages containing kernel init code, which are released after kernel init is completed)
```
paolonig, line 86: by monolithic portion you mean code compiled in the Image and not kmod'd later?

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+      1. the kernel doesn't spontaneously repurpose the utilisation of its own physical pages; therefore it is possible to assume that the logical content of kernel allocations will reamin tied to the associated physical pages, as long as it is not intentionally altered (or subject to interference)
+         1. metadata memory used by the kernel for housekeeping purposes related to processes is included in this category; examples: task/cred structures, vmas, maple leaves, process page tables.
+      2. memory used by the kernel for the actual userspace processes: the content of this logical page determines its life cycle and expectancy: certain content such as code or constants can be "re-constructed" by relaoding it from file (code pages are likely to necessitate undergoing re-linking), so the actual logical content might disappear, over time. Other pages, on the other hand, are meant to hold non-re-constructible content, such as stack, heap, and variable data. These pages can, at most, be swapped out, and loaded back later on, but they cannot be simply dropped.
```
paolonig, line 97: can you better explain what you mean by "the actual logical content might disappear, over time" ?

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: At least on ARM64 the access through user space mappings is disabled, unless TTBR0 is accidentally enabled (and that would count as 1 interference); to get interference toward the content of the userspace process memory, a rogue operation toward user-space address range shuold happen (and that would count as 2nd interference).

So, my reasoning is that if we consider only 1 interference event, then it is not possible.

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: I do not know how to express it, but sequence is not referred to the temporal sequence, as it seems you read this.
It means that pages in a process memory are (usually) only virtually linear. Therefore the kernel needs to use the process memory map, to know the correct order of pages. Imagine you have a structure that is placed across a page boundary, in user memory. The kernel cannot know which ones are the 2 adjacent pages, without using the process memory map.
And when it does, it is able to pretent being a process, unless there is something preventing it. Like PXN and PAN.

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: "very often" doesn't mean "always"; besides, the scenario where the kernel writes directly already exists, and it is done by using the user space mappings through TTBR0.
The DMA is relevant because it completely bypasses the MMU

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+      2. the kernel also keeps ready a certain amount of pre-diced memory allocations, to avoid incurring the penalty of having to look for some free memory as a consequence of an allocation request.
+      3. folios are structures introduced to simplify the management of what has been traditionally called compound pages: a compound page represents a group of contiguous pages that is treated as a single logical unit. Folios could eventually support the use of optimisations provided by certain pages (e.g. ARM64 allows the use of a single page table entry to represent 16 pages, as long as they are physically contiguous and aligned to a 16 pages boundary, through the "contiguous bit" flag in the page table). This can be useful e.g. when keeping in the page cache a chunk of data from a file, should the memory be released, it could result in releasing several physically contiguous pages, instead of scattered ones.
+   6. whenever possible, allocations happen through caches, which means that said caches must be re-filled, whenever they hit a low watermark, and this re-filling can happen in two ways:
```
paolonig, line 110: above by caches do you mean "page cache" as described above? if so maybe we should use this name to be consistent....

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+   10. The Linux kernel provides means to limit certain requests a process might present; for example with cgroups it is possible to create logical memory "bubbles" that cannot grow beyond a set size, and associate processes to them, that share the collective limit within the bubble. But this does not much toward separating how the kernel might use the underlying memory, besides setting the constraint as described.
+      
+#### **Safety-Oriented consideration**
+The following considerations are of a more deductive nature.
+
+1. Because of the way pages, fractions and multiples of them are allocated, freed, cached, recovered, there is a complex interaction between system components at various layers.
+2. Even using cgroups, it is not possible to segregate interaction at the low level between components with differnet level of safety qualification (e.g. a QM container can and most likely will affect the recirculation of pages related to an ASIL one)
```
paolonig, line 122: Above I do not think that the issue is a QM container affecting the recirculation of pages used by an ASIL one but rather the current recirculation mechanism not distinguishing between pages to be used by QM and ASIL containers. If so let's phrase this in this way, because if the recirculation mechanism works as expected there would be no impact from a misbehaving QM workload

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+3. Because of the nature of memory management, it must be expected that memory management mechanisms will  interfere with safe processes, either due to a bug or due to the interference toward the metadata they rely on. For example, the memory management might hand over to a requesting entity a memory page that is currently already in use either by a device driver or by a userspace process playing a role in a safety use case.
```
paolonig, line 123: I do not agree on the sentence <<it must be expected that memory management mechanisms will  interfere>> as it sounds that you have enough evidence to claim that it is buggy.
I would say instead <<Because of the nature of memory management, it must be noted that an eventual bug in the memory management mechanisms or an interference toward the metadata they rely on is likely to turn into a hazard for the running safety workloads>>

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: I am not describing the design mechanisms only. My goal is really to present together all the routes, so that even this is not underestimated.

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: yes, if here is a better term, I do not know it.

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: "live content" ? 
The content as it is intended to be in memory when the process needs to use it.

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: not just that - you have also all the metadata used by the kernel for managing pages meant to be used to support processes

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: I have evidence to prove that it is exposed to each and every bug that might be present in the kernel, because all the metadata in use is not write protected.
So it's up to whoever wants to claim that this corruption is not possible, to prove it.
Even assuming that somehow a memory manager had been qualified as ASILB or even above, but out of context, it is the cascaded interference throguh the metadata that voids that qualification.

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+4. Still due to te complex interaction between processes, kernel drivers and other kernel code, it is practically impossible to qualify the kernel as safe through positive testing alone, because it is impossible to validate all the possible combinations, and it is equally impossible to assess the overall test coverage and the risk associated with not reaching 100%. The only reliable way to test is to use negative testing (simulating a certain type of interference) and confirming that the system response is consistent with expectations (e.g detect the interference, in case of ASILB requirements). And even then, the only credible claim that can be made is that, given the simulated type of interference, on the typology of target employed, the reaction is aligned with the requirements. Other types of targets will require further ad-hoc negative testing.
```
paolonig, line 124: There are few points that are not clear above:
1) Why you are associating positive testing with "all the possible combinations" (the expectations is to cover the appropriate equivalence classes)
2) Your definition of negative testing should be further contextualised. For example I could interpret negative testing as in testing the behavior of kmalloc in case of error (e.g. in case no memory is available)

In general I do not agree with your statement that mix the MM systematic capability claim and the interference from other Kernel components. Also I do not like the fact that you impose only runtime safety measures as the only way to claim FFI within the Kernel.

what I would say is that:
a) the memory management subsystem is very complex and defining a set of FuSa qualification activities (including comprehensive testing) to achieve a target systematic capability is very costly
b) Even if a systematic capability claim is achieved on the memory management, the same does not cover interference from other components of the Kernel. In order to claim freedom from such interferences dedicated runtime safety measures and/or dedicated verification activities shall be defined.
The verification or runtime safety measures shall be tested by injecting interference modes and testing the their effectiveness

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+5. Linux Kernel mechanisms like SELinux and cgroups/containers do not offer any protection against interference originating from the kernel itself.
+6. The Linux Kernel must be assumed to be QM, unless specific subsystems are qualified through both positive and negative testing.
```
paolonig, line 126: Above I would say "Kernel code must be assumed NSR (non safety related), unless the integrator performs additional verification activities to claim a specific systematic capability target (including QM)"

Do we have upstream evidences of a QMS claim ?

igor-stoppa: I have been overenthusiastic :-D 
The problem I see with the proposed change is that it waters down the statement excessively.
I can reword it as "... must be assumed to not be safe; possibly QM at best ..."
But "veriication" is too bland because it becomes vvery subjective, while the point here is to stress he need for both positive and negative testing.

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+1. a process that is supposed to support safety requirements should  not have pages swapped out / dropped / missing,
+   because this would introduce:
+   1. uncertainty in the timing required to recover the content, if not immediately available
+   2. additional risk, involving the userspace paging mechanisms in the fulfilling of the safety requirements
+   3. additional dependency on runtime linking, in case the process requires it, and code pages have been
+      discarded - reloading them from disk will not be sufficient
+2. The optimisations made by the kernel in providing physical backing to process memory make it very
+   questionable if it can be assessed when a (part of) a process memory content is actually present in the
+   system physical memory.
+3. by default, it is to be expected that a process will be exposed to various types of interference from the kernel:
```
paolonig, line 207: Above I disagree on using the term "by default". Instead I would say "Unless the Kernel is qualified against a non-interence claim, it is to be expected ...."

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+   1. some of a more bening nature, like dropping of pages, or not allocation of not-yet-used one
+   2. some limited in extent, but hard or even practicaly impossible to detect, like a rogue write to process physical memory
+   3. some of systemic nature, like some form of use-after free, where a process page is accidentally in use also by another component
+   4. some of indirect nature, like for example when the page table of the process address space is somehow corrupted
+4. again, because of the extremely complex nature of the system, positive testing is not sufficient, but it needs to
```
paolonig, line 212: Above I would say "if a non-interference claim cannot be made, due to the extremely complex ...."

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+   be paired also with negative testing, proving that it is possible to cope with interference and detect it, somehow.
+5. the same considerations made about integrity vs. avaialbility for the kernel are valid here too: detecting
+   interference doesn't help with keeping it under a certain threshold, and due to the complexity of the system,
+   it is not possible to estimate the risk reliably.
```
paolonig, line 216: Estimate which risk? Availability risk ?

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+6. a single-thread process can interfere with itself, since typically most of its data is writable
```
paolonig, line 217: The statement above does not provide any info IMO. an ASIL B process running ASIL B code cannot interfere with itself due to the fact the the ASIL B code is qualified up to ASIL B, hence also any inteference bug is unlikely to happen (up to ASIL B certainty )

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+7. when dealing with a multi-threaded process, besides simple self interference, one must also consider cross-thread
```
paolonig, line 218: Above I would clarify that you are talking about a mutithread process where different threads run code of different systematic capability...

Contributions/Linux_Memory_Management_Essentials.md
paolonig: I would just add a note saying that we are not considering any swapping or any other memory management mechanism involved along PM routines....

Contributions/Linux_Memory_Management_Essentials.md
paolonig: but in the disclaimer above you wrote "When referring to specific HW features, the document refers to the ARM64 architecture" ....if you want to stay generic then we should remove this from the disclaimer.... ?

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: why different? it can be the same, even

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: exactly, up to ASILB, which means that it can still do it -  certainly less likely than if it were QM, but it is not ruled out, isn't it?

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: risk of lack of availability, but yes

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: I intentionally made a stronger statement.
Also to make it easier to prove me wrong :-)

Contributions/Linux_Memory_Management_Essentials.md
paolonig: Ah I see now.
Then I would say that in the context of the Kernel functions appointed to access user space pages, the Kernel uses the same translation mechanisms used by user space processes to access the physical memory

Contributions/Linux_Memory_Management_Essentials.md
paolonig: mmm then above maybe we could revisit the statement: <<5. the kernel can access userspace memory in three ways>> as <<5. the kernel can access userspace memory (by design or accidentally) in three ways>>

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: Sure you can disagree, but what evidence is there that the kernel will not interfere?
Qualification is a bit misleading though, because it is not clear on which ground.
You can have a "qualified" kernel and still incur in interference, because it was not proven that it cannot interfere, and the "qualification" happens to be incorrect.

"Qualified according to ISO26262" would be more acceptable, for example.

But here I'm really trying to stick to objective evidence.

A piece of paper will not stop a bug - the MMU (for example) might.

Contributions/Linux_Memory_Management_Essentials.md
paolonig: I would just say <<kernel code built as part of the boot Image>>, the reason being that the word monolithic is associated with the Kernel address space (on which KOs make no difference....)

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+5. Linux Kernel mechanisms like SELinux and cgroups/containers do not offer any protection against interference originating from the kernel itself.
+6. The Linux Kernel must be assumed to be QM, unless specific subsystems are qualified through both positive and negative testing.
+7. Even after achieving the ability to make certain claims about the Kernel integrity (or detection of its loss), this doesn't guarrantees system availability: knowing that an interference has occurred doesn't help with ensuring a certain level of system availability.
```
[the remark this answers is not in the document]
igor-stoppa: Eh, sadly no, it's not like that.
Even conceiving a kernel that for some reason is able to reject interference from a safety standpoint, that only means that those parts which are exposed to hazards will not be corrupted.

That is fine and dandy. But what about the rest? E.g. is the task structure of a task protected? Because in case it is not, it is exposed to interference as well, adn the task might not be run timely.

Contributions/Linux_Memory_Management_Essentials.md
paolonig: Ah ok I see...you meant to say that "this type of content is safe to disappear, over time" (since it can be reconstructed) ?

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: No, no, I agreed. I was explaining why I initially omitted this. But I can add it.

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: 1. equivalence classes tend to be subjective, and prone to errors. If you go down the path of "equivalence classes" you have to prove both correctness and completeness of these. Which seems intreatable.
2. if I am talking about interference, isn't it clear that it erfers to whatever interference is being analysed?


I'm sorry you do not like my statement about runtime FFI.
Perhaps you can prove that other approaches will yield the same results?

Regarding "costly" I'm not sure if that is the right word, because it implies that one might achieve the result, spending enough. But even if money was not a concern, time would make it impossible.

And I do not beleive that verification activity will be enough, because you always have the problem of proving completeness.

If you cannot say what has been left out, how can you say that what was done is sufficient?

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: from the perspective of a newcomer it might be useful to add maybe more words - I'll think about some compromise

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: ok

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: ok

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: ok

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: no, I was actually pointing to the risk associated with it disappearing and possiby not coming back correctly and/or timely XD 

but I could say oemthing like "by design this content miht disappear, and is expected to be reconstructed if/when needed" and then point to the associated risk

dweingaertner: I think it helps splitting this paragraph into the two separate cases: a) reconstructible and b) live or non-reconstructible pages

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+      to represent optimised contiguous pages on ARM64 (instead of mapping 16 entries, it is possible
+      to map a 16-pages chunk of physically contiguous memory, that is also aligned to a 16 pages
+      boundary, reducing TLB use).
+      1. A folio acts as intermediary between vma and lower level memory management
+      2. it might pre-allocate/map more pages than explicitly requested
+7. read-ahead: when asked to fetch data from disk, the kernel might attempt to optimise the operation,
+   reading more pages than requested, under the assumption that more requests might be coming soon
+ 
+
+#### **Safety-Oriented consideration**
```
dweingaertner, line 195: Same comment as above. We should to separate the premises from these considerations, and explicitly state the premises each conclusion builds on. Take for example the 1 argument below. Uncertainty in the timing might be a problem, but might not, depending on the requirements or on the size of the uncertainty.

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+      2. the kernel also keeps ready a certain amount of pre-diced memory allocations, to avoid incurring the penalty of having to look for some free memory as a consequence of an allocation request.
+      3. folios are structures introduced to simplify the management of what has been traditionally called compound pages: a compound page represents a group of contiguous pages that is treated as a single logical unit. Folios could eventually support the use of optimisations provided by certain pages (e.g. ARM64 allows the use of a single page table entry to represent 16 pages, as long as they are physically contiguous and aligned to a 16 pages boundary, through the "contiguous bit" flag in the page table). This can be useful e.g. when keeping in the page cache a chunk of data from a file, should the memory be released, it could result in releasing several physically contiguous pages, instead of scattered ones.
```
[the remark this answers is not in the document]
igor-stoppa: yes, but please understand that I'm not trying to explain everything that is associated with memory management, or else this will becoem a book in itself - I expect the reader to do own research on topics that might not be fully known. Or others to contribute references. The topic is complex and one who wants to work on it is expecccted to study it.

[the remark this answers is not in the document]
igor-stoppa: anyways, in this specific case the purpose of the contiguous bit it to use only 1 TLB entry instead of 16 - but then I shuld explain what is a TLB and so on ...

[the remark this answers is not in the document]
dweingaertner: The thing is not that you have to explain further what the contiguous bit does, since this is clear. I just cannot see why go into that amount of detail about the implementation of folios. I think this over-complicates the whole paragraph. And then the last sentence feels detached from the beginning.
In my opinion, the important thing to know about folios is that they represent more than one physical page, and might be used to optimize memory access. But if you go deeper, and argue something is useful, then I think it is better to state clearly why, and not let the reader guessing.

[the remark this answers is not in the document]
igor-stoppa: Ah, I see. Well, avoiding memory fragmentation is the goal, but I kind of took it for granted that it would stand out. I can add this.

Contributions/Linux_Memory_Management_Essentials.md
reiterative: I agree that power management considerations could be a separate point, but we could perhaps make the following changes to clarify this point: 

```suggestion
1. Unlike processes memory, kernel memory pages are not dynamically swapped during normal (i.e. not low-power) operation, nor dropped silently by the kernel itself,
```

The point about hypervisors could then be extended to note that memory may be swapped when entering or leaving low-power states. It might be better to break this out as a separate sentence.

Contributions/Linux_Memory_Management_Essentials.md
reiterative: Agreed. Please reference PXN / PAN as examples.

Contributions/Linux_Memory_Management_Essentials.md
reiterative:
```suggestion
      4. the kernel is able to access the pages of a user space process using the same translation mechanisms as the process (i.e. the process memory map), unless explicitly prevented by e.g. PAN or PXN
```

Contributions/Linux_Memory_Management_Essentials.md
reiterative: This is a valid qualification, but perhaps it could be reworded like this:

```suggestion
      3. in many cases, the kernel does not access this area directly, but rather configures a DMA controller to complete the transfer directly to the physical memory (i.e. bypassing the MMU)
```

Contributions/Linux_Memory_Management_Essentials.md
reiterative: Would accessing via the EL1 mappings and linear map represent a deliberate bypass (and thus arguably misuse) of an unspecified design feature? If it is a conceivable approach that a kernel component might use (legitimately or otherwise), then we should consider it as a potential mechanism.

Contributions/Linux_Memory_Management_Essentials.md
reiterative: Agreed

Contributions/Linux_Memory_Management_Essentials.md
reiterative: +1 for breaking this up into two points

Contributions/Linux_Memory_Management_Essentials.md
reiterative: Could we add a list of the applicable caches in brackets to clarify this?

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: no, it's not an abuse, that is how copy operations to/from userspace work
it is perfectly normal and justifiable

Contributions/Linux_Memory_Management_Essentials.md
reiterative:
```suggestion
2. Even using cgroups, it is not possible to eliminate indirect interaction at the low level between components with different levels of safety integrity (e.g. the recirculation of pages related to critical processes in one group might be affected by less critical processes in another group)
```

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: done

Contributions/Linux_Memory_Management_Essentials.md
reiterative:
```suggestion
3. Because of the nature of memory management, we cannot rule out the possibility that memory management mechanisms will interfere with safe processes, either due to a bug or due to the interference toward the metadata they rely on. For example, the memory management might hand over to a requesting entity a memory page that is currently already in use either by a device driver or by a userspace process playing a role in a safety use case.
```

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: done

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: done

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: I think the division done by @dweingaertner is not correct; it is all somehow reconstructible - if one allows for swap. Preventing swap simply means that only what is in theory reconstructible gets dropped. But the problems with the reconstructions apply.

Contributions/Linux_Memory_Management_Essentials.md
reiterative: I think the main substance of this point can be made without some of the potentially contentious references. What about: 

```suggestion
4. These complex interaction between processes, kernel drivers and other kernel code mean that it is practically impossible to qualify the kernel through positive testing alone, because the specification task and set of required test cases would be enormous.  
   1. We can reduce the scope by specifying a kernel version, configuration, target system, software integration, and set of application use cases, but the number of required test cases would still be very large. 
   2. However, the use of negative tests in combination with positive tests could make a verification-based approach viable for a specific integration, target system and use case, by specifying a set of failure modes for which the behaviour has been verified. 
   3. This might be achieved by e.g. simulating an identified type of interference for a range of positive test cases, and confirming that the overall system response is consistent with specified expectations (e.g interference is detected and a mitigation is triggered).
```

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: explained in the amended text

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: In this case "enormous" is not enormous enough :-D 
Only negative testing provides valid evidence, because using positive testing alone is equivalent to claim that the code is devoid of a class of bugs (interference bugs).

With negative testing, it is easy to verify whatever claim that might be made.
Without it, there is a completeness problem.

Whoever thinks they can provecompleteness (or even give a verifiable measurement of confidence, like "reliable at 95% because of X, Y and Z") is welcome to come forward.

Till then, the completeness problem remains unsolved.

Contributions/Linux_Memory_Management_Essentials.md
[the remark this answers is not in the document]
igor-stoppa: added

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: done

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+   8. freeing of memory pages also happens in a deferred way, through separate threads, so that there is no overhead on the freer, in updating the metadata associated with the memory that has just been released.
+   9. All of the mechanisms described above for memory management are not only memory providers, but memory users as well, because they rely on metadata that cannot be pre-allocated and must be adjusted accordingly to the memory transactions happening, as the system evolves over time; therefore they also consume memory, generating an overhead, and being exposed themselves to interference.
+   10. The Linux kernel provides means to limit certain requests a process might present; for example with cgroups it is possible to create logical memory "bubbles" that cannot grow beyond a set size, and associate processes to them, that share the collective limit within the bubble. But this does not much toward separating how the kernel might use the underlying memory, besides setting the constraint as described.
+      
+#### **Safety-Oriented consideration**
```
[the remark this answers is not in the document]
igor-stoppa: i'm open to debate on this idea, but I would prefer to first exhaust the comments gotten so far, by keeping this as one single doc

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: @reiterative if you want to have anohter stab at it, while keeping the points I listed in the previous comment, we might be close to a resolution.

reiterative: Second attempt: 
```suggestion
4. These complex interaction between processes, kernel drivers and other kernel code mean that it is practically impossible to qualify the kernel through positive testing alone.
   1. Specifying requirements and implementing a credible set of tests to cover all of the kernel's functions for the general case is certainly infeasible (and arguably impossible), because the range of potential applications and integrations for Linux is too broad.     
   2. We can constrain this scope by specifying a kernel version and configuration, for a given set of target systems and software integrations, and specifying the set of functions it is intended to provide. However, this would still not be sufficient to assert that the kernel is devoid of certain classes of bug (e.g. for bugs caused by interference).     
   3. Negative tests derived from credible analysis of the kernel could be used to address this, by verifying its behaviour (and/or mitigations provided by other components of a target system) for a documented set of failure modes. 
   4. This might be achieved, for example, by simulating an identified type of interference for a range of positive test cases, and confirming that the overall integrated system's response is consistent with a specified set of expectations (e.g. the interference is detected and a kernel- or system-level mitigation is triggered).
   5. This, in combination with requirements-based functional testing, could be a viable approach for qualifying a specific integration and configuration of Linux, for a given set of target systems and use cases.
```

Contributions/Linux_Memory_Management_Essentials.md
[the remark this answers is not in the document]
reiterative:
```suggestion
7. Claims about kernel integrity (or detection of its loss), do not guarantee system availability; safety arguments for a Linux-based system that rely upon a level of availability must separately show that this is supported.
```

Contributions/Linux_Memory_Management_Essentials.md
reiterative:
```suggestion
3. we cannot rule out the possibility that a process will be exposed to various types of interference from the kernel:
```

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+   because this would introduce:
+   1. uncertainty in the timing required to recover the content, if not immediately available
+   2. additional risk, involving the userspace paging mechanisms in the fulfilling of the safety requirements
+   3. additional dependency on runtime linking, in case the process requires it, and code pages have been
+      discarded - reloading them from disk will not be sufficient
+2. The optimisations made by the kernel in providing physical backing to process memory make it very
+   questionable if it can be assessed when a (part of) a process memory content is actually present in the
+   system physical memory.
+3. by default, it is to be expected that a process will be exposed to various types of interference from the kernel:
+   1. some of a more bening nature, like dropping of pages, or not allocation of not-yet-used one
```
reiterative, line 210: I don't understand what 'not allocation of a not-yet-used one' means. 

```suggestion
   1. some of a more benign nature, like dropping of pages, or not allocation of not-yet-used one
```

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,249 @@
...
+   2. some limited in extent, but hard or even practicaly impossible to detect, like a rogue write to process physical memory
+   3. some of systemic nature, like some form of use-after free, where a process page is accidentally in use also by another component
+   4. some of indirect nature, like for example when the page table of the process address space is somehow corrupted
+4. again, because of the extremely complex nature of the system, positive testing is not sufficient, but it needs to
+   be paired also with negative testing, proving that it is possible to cope with interference and detect it, somehow.
+5. the same considerations made about integrity vs. avaialbility for the kernel are valid here too: detecting
```
reiterative, line 216:
```suggestion
5. the same considerations made about integrity vs. availability for the kernel are valid here too: detecting
```

Contributions/Linux_Memory_Management_Essentials.md
reiterative, line 218:
```suggestion
   it is not possible to reliably estimate the risk that availability criteria will not be achieved.
```

Contributions/Linux_Memory_Management_Essentials.md
reiterative: Perhaps clarify that this is a risk the kernel is not able to mitigate? Or just drop this point

 ```suggestion
6. a single-thread process can interfere with itself, since typically most of its data is writeable; the kernel cannot be responsible for preventing this category of interference   
```

File: Contributions/Linux_Memory_Management_Essentials.md (hunk 18 above)
Comment (reiterative) [reply Contributions/Linux_Memory_Management_Essentials.md:220]: Agreed: this is describing the mechanism of interference, which is independent of systematic capability considerations

File: Contributions/Linux_Memory_Management_Essentials.md (hunk 12 above)
Comment (igor-stoppa) [reply Contributions/Linux_Memory_Management_Essentials.md:126]: the whole point is to say that it must be possible to confirm that the claim is correct

given the nature of the problem (one or more classes of bugs), it is not realistic to say that positive testing can cover them all, period, it is the same as claiming that the code is bug free, for those classes

the system subject to testing can comprise also components external to the kernel, but there needs to be something that can prove kernel errors will be detected, or else we are saying that we think it is possible to remove completely certain classes of errors form the kernel, which would be preposterous

File: Contributions/Linux_Memory_Management_Essentials.md (hunk 13 above)
Comment (igor-stoppa) [reply Contributions/Linux_Memory_Management_Essentials.md]: I made the change, so closing this as obsolete

File: Contributions/Linux_Memory_Management_Essentials.md (hunk 20 above)
Comment (igor-stoppa) [reply Contributions/Linux_Memory_Management_Essentials.md:197]: Agreed to reconsider later the overall posture of the document.
My motivation is to try to keep together statements which are kind of obvious with others that might be less obvious, but derivable.
We could consider - in general - to introduce also the notion of  which aspects are likely to be unavoidable, and which ones might be relevant only for some use case scenarios.

File: Contributions/Linux_Memory_Management_Essentials.md (hunk 16 above)
Comment (igor-stoppa) [reply Contributions/Linux_Memory_Management_Essentials.md]: closing as outdated

File: Contributions/Linux_Memory_Management_Essentials.md (hunk 12 above)
Comment (reiterative) [reply Contributions/Linux_Memory_Management_Essentials.md:126]: Third attempt: 
```suggestion
4. These complex interaction between processes, kernel drivers and other kernel code mean that it is not feasible to qualify the kernel through positive, requirement-based testing alone. 
   1. Safety-related claims made for the kernel must be specified and verified with a credible set of tests, which must necessarily include negative tests to demonstrate the effectiveness of any detection, mitigation and exception-handling mechanisms provided in support of these claims. 
   2. Credible analysis of the kernel could be used to devise a suitable set of tests for this, to verify the kernel's behaviour (and/or mitigations provided by other components of a target system) for a documented set of failure modes.
   3. Specifying requirements and implementing a credible set of tests to cover all of the kernel's functions for the general case is certainly infeasible (and arguably impossible), because the range of potential applications and integrations for Linux is too broad.
   4. Demonstrating that claims are supported with a combination of analysis and testing does not allow us to argue that the kernel is devoid of a certain class of bug, only that our verification mechanisms are sufficient to detect bugs affecting the specified claims, and for the identified failure modes.
   5.. Claims must therefore be verified for each iteration of the kernel intended for use in a safety-related context, and for the specific configuration and system integration of a target context; the sufficiency of the set of failure modes considered, and of the analysis informing these, must be also confirmed for the target context.
```

reiterative: Note that, while I'm happy to keep iterating on this, it might also be prudent to consider doing as [Daniel suggested](https://github.com/elisa-tech/wg-osep/pull/38#discussion_r1693042147) and land the safety considerations as a distinct set of changes from the (less contentious) technical analysis...

Contributions/Linux_Memory_Management_Essentials.md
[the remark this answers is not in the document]
igor-stoppa: I consider this resolved for now

Contributions/Linux_Memory_Management_Essentials.md
igor-stoppa: time flies ... but positive testing is still insufficient, at least witout evidence of the exhaustiveness of its coverage

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,261 @@
...
+2. The optimisations made by the kernel in providing physical backing to process memory make it very
+   questionable if it can be assessed when a (part of) a process memory content is actually present in the
+   system physical memory.
+3. we cannot rule out the possibility that a process will be exposed to various types of interference from the kernel:
+   1. some of a more benign nature, like dropping of pages, or not allocation of not-yet-used one
+   2. some limited in extent, but hard or even practicaly impossible to detect, like a rogue write to process physical memory
+   3. some of systemic nature, like some form of use-after free, where a process page is accidentally in use also by another component
+   4. some of indirect nature, like for example when the page table of the process address space is somehow corrupted
+4. again, because of the extremely complex nature of the system, positive testing is not sufficient, but it needs to
+   be paired also with negative testing, proving that it is possible to cope with interference and detect it, somehow.
```
reiterative, line 225: Is this slight re-framing sufficiently strong?

```suggestion
4. Because of the extremely complex nature of the system, positive testing alone is not sufficient: it must also be
   paired with negative testing, to prove that it is possible to detect or cope with given types of interference.
```

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,261 @@
...
+
+#### **Safety-Oriented consideration**
+The following considerations are of a more deductive nature.
+
+1. a process that is supposed to support safety requirements should  not have pages swapped out / dropped / missing,
+   because this would introduce:
+   1. uncertainty in the timing required to recover the content, if not immediately available
+   2. additional risk, involving the userspace paging mechanisms in the fulfilling of the safety requirements
+   3. additional dependency on runtime linking, in case the process requires it, and code pages have been
+      discarded - reloading them from disk will not be sufficient
```
reiterative, line 215: I hope this re-framing makes the applicable premises about safety-critical requirements clearer. As @dweingaertner suggests, it would make your deductive reasoning easier to follow if you set out some of the key premises / assumptions about what requirements / dependencies our hypothetical safety-critical processes are expected to have on the OS.

```suggestion
1. For a process that is supposed to support safety requirements, it may be desirable to avoid having pages 
   swapped out / dropped / missing, because this would introduce:
   1. Uncertainty in the timing required to recover the content, if not immediately available.
   2. Additional risk, involving the userspace paging mechanisms, in the fulfilment of applicable safety requirements
   3. Additional dependency on runtime linking, in cases where the process requires it: where code pages have 
      been discarded, reloading them from disk may not be sufficient to meet applicable timing requirements.
```

igor-stoppa: @reiterative I would remove completely should/may/would and then take a more matter-of fact approach.
And then just state that having pages swapped out/dropped/missing DOES introduce those problems.
It is undeniable and easily demonstrable.

It can then be follwed by an equally matter-of-fact statement that since these problems are very real, they must be addressed.

I think nobody would disagree to this.

reiterative: This makes sense: I've updated my suggestion.

One question about this: how is point 3 (about 'runtime linking') different to points 1 and 2?

reiterative: Thanks, that makes more sense to me now. I've changed _runtime linking_ to _the runtime linker_ in my suggestion.

reiterative: But you've already updated the text, so you can ignore.

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,261 @@
...
+The following considerations are of a more deductive nature.
+
+1. Because of the way pages, fractions and multiples of them are allocated, freed, cached, recovered, there is a complex interaction between system components at various layers.
+2. Even using cgroups, it is not possible to eliminate indirect interaction at the low level between components with different levels of safety integrity (e.g. the recirculation of pages related to critical processes in one group might be affected by less critical processes in another group)
+3. Because of the nature of memory management, we cannot rule out the possibility that memory management mechanisms will interfere with safe processes, either due to a bug or due to the interference toward the metadata they rely on. For example, the memory management might hand over to a requesting entity a memory page that is currently already in use either by a device driver or by a userspace process playing a role in a safety use case.
+4. These complex interaction between processes, kernel drivers and other kernel code mean that it is not feasible to qualify the kernel through positive, requirement-based testing alone.
```
reiterative, line 126:
```suggestion
4. These complex interactions between processes, kernel drivers and other kernel code mean that it is not feasible to qualify the kernel through positive, requirement-based testing alone. 
```

Contributions/Linux_Memory_Management_Essentials.md
```diff
@@ -0,0 +1,261 @@
...
+1. For a process intended to support safety requirements, having pages swapped out, dropped or missing
+   creates additional risk, because it introduces:
+   1. Uncertainty in the timing required to recover the content, if it is not immediately available.
+   2. Reliance on userspace paging mechanisms for the fulfilment of applicable safety requirements
+   3. Additional dependency on runtime linking: where code pages have been discarded, reloading 
+      them from disk can cause a process to violate its applicable timing requirements.
+2. The optimisations made by the kernel in providing physical backing to process memory make it very
+   questionable if it can be assessed when a (part of) a process memory content is actually present in the
+   system physical memory.
+3. we cannot rule out the possibility that a process will be exposed to various types of interference from the kernel:
```
reiterative, line 219:
```suggestion
3. We cannot rule out the possibility that a process will be exposed to various types of interference from the kernel:
```

Contributions/Linux_Memory_Management_Essentials.md
reiterative: I agree that the structure of this document would need some further thought before publishing it, but I would prefer to merge it as a Contribution in its current form, and then consider how best to restructure it for publication (to https://osep.elisa.tech/) in a separate MR.

PR #51, opened 2025-01-23
Detailing SELinux,cgroups pros nad cons vs Hypervisor.

igor-stoppa:
As discussed, I've added some extra text.
It looked very weird to have it at the beginning, though, when it came to add it.
So I preferred to just queue it to the original text being debated.

### DIFF
--- a/docs/Interference_Scenarios_for_an_ARM64_Linux_System.md
+++ b/docs/Interference_Scenarios_for_an_ARM64_Linux_System.md
@@ -472,8 +472,8 @@ This is perfectly fine from the perspective of containing user-space, however, s
 frequency execution of a large amount of code which can cause either direct or indirect interference.
 And such interference is not always detectable, depending on which component it might affect.
 
-These means are usualy enabled in a mixed criticality scenario, when attempting to contain interference.
-The choice of enabling them, despite the associated risk, might be driven by overall considerations about choosingthe lesser evil.
+These features are usually enabled in a mixed criticality scenario, when attempting to contain interference.
+The choice of enabling them, despite the associated risk, might be driven by overall considerations about choosing the lesser evil.
 Of course one could attemtp to qualify them, but then it is necessary to consider the fact that in reality it is necessary to qualify them together with the user-space-provided policies they will enact.
 Without being configured by user-space, neither SELinux nor cgroups are of any particular use.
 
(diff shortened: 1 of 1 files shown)


docs/Interference_Scenarios_for_an_ARM64_Linux_System.md
```diff
@@ -472,6 +472,16 @@ This is perfectly fine from the perspective of containing user-space, however, s
 frequency execution of a large amount of code which can cause either direct or indirect interference.
 And such interference is not always detectable, depending on which component it might affect.
 
+These means are usualy enabled in a mixed criticality scenario, when attempting to contain interference.
+The choice of enabling them, despite the associated risk, might be driven by overall considerations about choosingthe lesser evil.
```
reiterative, line 476:
```suggestion
These features are usually enabled in a mixed criticality scenario, when attempting to contain interference.
The choice of enabling them, despite the associated risk, might be driven by overall considerations about choosing the lesser evil.
```

docs/Interference_Scenarios_for_an_ARM64_Linux_System.md
```diff
@@ -472,6 +472,16 @@ This is perfectly fine from the perspective of containing user-space, however, s
...
+Of course one could attemtp to qualify them, but then it is necessary to consider the fact that in reality it is necessary to qualify them together with the user-space-provided policies they will enact.
+Without being configured by user-space, neither SELinux nor cgroups are of any particular use.
+
+An alternative - possibly more costly - path could be to instead isolate more safety relevant loads from non safety relevant ones, introducing a second virtual machine, with an hypervisor underneath.
```
reiterative, line 480:
```suggestion
An alternative - possibly more costly - path could be to instead isolate more safety-relevant loads from non-safety-relevant ones, introducing a second virtual machine, with a hypervisor underneath.
```

docs/Interference_Scenarios_for_an_ARM64_Linux_System.md
```diff
@@ -472,6 +472,16 @@ This is perfectly fine from the perspective of containing user-space, however, s
...
+The caveat is that now the hypervisor can be a source of interference. And it is also necessary to have HW capable to support an EL2.
+It can be an interesting alternative, though, if using a Type1 hypervisor (like Xen), because it is relatively simpler than trying to qualify the Linux code.
```
reiterative, line 482:
```suggestion
It can be an interesting alternative, though, if using a Type1 hypervisor (like Xen), because it is relatively simple in comparison to qualifying the Linux code.
```

reiterative:
I've added some suggestions, but please resubmit this using a branch in the OSEP repository itself, as the Actions pipeline currently breaks when the originating branch is in a fork.

PR #59, opened 2026-07-01
Draft: Recommended approaches and rationale

reiterative:
Write up recommended approaches and rationale based on workshop presentation and previous discussions


docs/approaches.md
```diff
@@ -0,0 +1,115 @@
+# What approaches do we recommend?
+
+Our starting position in this working group is that Linux is **not viable as a
+safety-critical operating system in its own right**, for the simple reason that
+it has not been designed, implemented or verified with this purpose in mind.
+
+This means that _Enabling Linux in Safety Applications_ involves understanding
+the risks that using Linux entails, and then designing system architectures that
+can manage or mitigate those risks, with or without the involvement of Linux
+itself.
```
reiterative: This requires understanding the system in question from engineering first principles, in order to identify the risks and document how the system is intended to address these.

docs/approaches.md
```diff
@@ -0,0 +1,115 @@
...
+  code[^4]) and the very broad scope of its supported functionality, hardware
+  and configurations.
+- This means that such an approach would inevitably need to focus on a
+  tightly-constrained subset of supported functionalities and hardware, to
+  reduce the scope of the work required.
+- This would typically also require the removal of all extraneous or undesired
+  source code and configurations (and modification of the associated build files
+  and tests), to avoid having to analyse these and their associated risks, and
+  having to justify their inclusion in the software subset.
+- Any deficiencies identified by analysis of the chosen Linux version and subset
```
reiterative, line 64: The need to fix any deficiencies that you find is a disincentive if this results in more work for you to do.

reiterative (commented):
Comments collected in discussions in the OSEP weekly meeting

docs/approaches.md
```diff
@@ -0,0 +1,115 @@
+# What approaches do we recommend?
```
paolonig, line 1: Lighthouse WG feedback: 
1) The title should specify the intent of the approach (i.e. approach to do what?)
2) Who is "we"

Gab feedback: From my personal standpoint it is possible to fork linux and safety certify it as a SEooC (e.g. see RHIVOS certificates). So at least myself is not part of 'we' in this PR

docs/approaches.md
```diff
@@ -0,0 +1,115 @@
...
+
+Our starting position in this working group is that Linux is **not viable as a
```
paolonig: Lighthouse WG feedback: it is better to replace "Linux" with "Vanilla Linux" (i.e. we do not know if some integrators or OEMs have proven in use or SW systematic capability evidences that would be sufficient to do what you recommend not to do)

docs/approaches.md
```diff
@@ -0,0 +1,115 @@
...
+However, the criteria for applying this type of argument have very strict
+requirements relating to failure rates from use of the element in comparable
+products for the same purpose, based on incident report data collected over long
+periods of time[^1].
+
+Even if there were such data available for Linux, we could not make an argument
+based on this unless we used the same version of the software, in the same kind
+of system for the same purpose.
+
+## Why not fork Linux and safety certify it?
```
paolonig, line 42: Lighthouse WG Feedback: It should be specified that this  remark is valid for the current state of Vanilla Linux, given the technical debt that comes with it. However ELISA is working towards reducing such a debt  (e.g. in the Lighthouse WG as well as in the Safety Arch WG as well as in the Tool WG); one of ELISA's goal is to make this easier in the future (so that we do not need to consider architectural decomposition solutions only).

docs/approaches.md
```diff
@@ -0,0 +1,115 @@
...
+  Linux analysed, for the specific use case(s) considered and for whatever
+  constrained set of functions and target hardware was selected.
+- As a consequence, any fixes, enhancements or new features added in the
+  upstream Linux could not be automatically applied to the customised codebase,
+  instead requiring detailed impact analysis to understand their potential risks
+  and relevance.
+- Maintaining such a codebase would be an ongoing commitment for the lifetime of
+  any products making use of the special Linux
+
+## Why not show that Linux is safe by testing it thoroughly?
```
paolonig, line 77: Lighthouse WG feedback: The point here should be that increased testing is not sufficient if it is not traced to corresponding safety claims, safety requirements and safety scope of the code being tested and supported by a solid verification strategy.
This is quite a complex problem that cannot be summarised by a statement like <<All such measures have a dependency on specified requirements for the software, the system(s) within which it is intended to operate, and the safety-related functions that the system(s) must provide.>>.

So it is a good start but it should be structured better (maybe with some examples about what testing-only can lead to)

docs/approaches.md
reiterative: I've updated the title and text to clarify what is meant by 'we'.

docs/approaches.md
reiterative: I've added a section about this.

docs/approaches.md
reiterative: Agreed. I've added some of your text, and plan to extend this section to provide more detail.

docs/approaches.md
reiterative: I have updated the framing to try to make this clearer.

docs/approaches.md
reiterative: Text updated to reflect this.

docs/approaches.md
reiterative: Text updated to include this.

